PR #1, opened 2017-11-26
Use arduino i2c wire functions and increase i2c frequency to 400kHz. …

tedgreen99:
… Uses pins A4 (SDA) and A5 (SCL).

### DIFF
--- a/gbs-control.ino
+++ b/gbs-control.ino
@@ -942,6 +942,9 @@ void resetSyncLock() {
 
 void setup() {
   Wire.begin();
+  // The i2c wire library sets pullup resistors on by default. Disable this so that 5V arduinos aren't trying to drive the 3.3V bus.
+  digitalWrite(SCL, LOW);
+  digitalWrite(SDA, LOW);
   // TV5725 supports 400kHz
   Wire.setClock(400000);
 
--- a/gbs-control.ino
+++ b/gbs-control.ino
@@ -1,4 +1,4 @@
-#include "I2CBitBanger.h"
+#include <Wire.h>
 #include "ntsc_240p.h"
 #include "pal_240p.h"
 #include "hdtv.h"
@@ -6,9 +6,8 @@
 #include "test720p.h"
 #include "pal_288p.h"
 
-// bitbanger developed reversing the original firmware I2C timings and signaling by mybook4
-// https://github.com/mybook4/DigisparkSketches/tree/master/GBS_Control
-I2CBitBanger i2cObj(0x17); //GBS_I2C_ADDRESS  // 0x2E
+// 7 bit GBS I2C Address
+#define GBS_ADDR 0x17
 
 // I want runTimeOptions to be a global, for easier initial development.
 // Once we know which options are really required, this can be made more local.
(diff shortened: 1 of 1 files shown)


tedgreen99:
Hi Rama,  I posted a message on the shumps forum explaining the rational for this change but it needs to approved by an admin before it will show up.

ramapcsx2:
Hi, thanks for helping out! 

It took about a day for the admins at shumps to grant me access.

This looks like you're bypassing the bitbanger. 
I assumed the GBS used some form of odd timing or signaling, hence requiring it. I never tested the regular wiring library.

If this works, that'd be great! A standard library will be easier to port, for one, and it might even be faster with that 400khz setting. I don't know how fast the bitbanger works but it had some reserves when I tried reducing the delays. I could nearly remove them even.

I'll check this out! :)

ramapcsx2:
Yep, this works out of the box. And it's super fast, too :)

I did a quick safety check and there's one problem so far:
Apparently the default for the wiring library is to enable the pull-up resistors.
I hope this can be disabled, as it would break 5V Arduino on 3.3V GBS operation.
I'll check the docs :)

tedgreen99:
From what I could get out of the docs it seemed to imply that the i2c pins would follow the i2c spec and require external pull ups.  The example circuits have external pull ups.  I'll see if I can discern what the library is actually doing.

ramapcsx2:
`Wire.begin();
  digitalWrite(SCL, LOW);
  digitalWrite(SDA, LOW);
  Wire.setClock(400000);`
This works well here :)
The GBS has external 2k7 pull-up resistors, so the Arduino should never drive an I2C pin (or any pin) high.

tedgreen99:
I would be worried about a momentary current spike between the time from Wire.begin to the disabling of the pullups.
https://github.com/ninjablocks/arduino/blob/master/Wire/utility/twi.c
On line 69 it enables the pullups.
The Wire.begin is simple enough that we could use our own custom routine which doesn't enable the pullups.

tedgreen99:
Actually, after thinking about it, I don't think there would be much issue with the solution you suggested.  The current should be pretty minimal and only for a short period.

ramapcsx2:
On my analog scope, I couldn't even see the 5V blip. But the trigger fired, so it was there.

I like your idea of substituting twi_init().
There should be a way to override a library function. I remember reading about it somewhere.
I would then still do the manual pull-up disable, just to future proof this for when people start porting :)

tedgreen99:
It's only possible to override a C++ function if it is declared as virtual.  Unfortunately the twi_init is not virtual so it would be necessary to have our own custom wire.begin function and call it instead.

ramapcsx2:
Bummer. 
For now, I'm fine with just disabling the pull-ups after Wire.begin().
The 5V burst can only last for a few Arduino MCU cycles and it's likely that it doesn't even reach the 5275 I2C pins.
Each cycle is 62ns and it does like 7 cycles on calculation there + the function return.
The 2 lines have 33 Ohm in series and those 2k7 Ohm resistors to 3.3V (and the inputs of that original control processor).
Yeah, the burst won't even reach the resistors.

ramapcsx2:
So with no load connected, there is a 5V blip for 0.1ms, going to to 3.3V in 0.2ms.
With the wire connected, there's nothing anymore.
Safe! :)

PR #6, opened 2018-03-12
New sync locking mechanism

bkoropoff:
This is not 100% ready, but I want to get feedback on it.  It prevents tearing artifacts with my Sega Genesis.

------------------------

Use new sync locking technique

This requires a hardware change: the input csync signal must be
connected to an analog input pin.

The input and output vsync pulses are phase locked by constantly
adjusting the size of the vertical front porch.  The output is locked
to trail the input so that it never "catches up" with the capture
buffer being filled by the deinterlacer, which would result in a
screen tear.

### DIFF
--- a/gbs-control.ino
+++ b/gbs-control.ino
@@ -1357,15 +1357,15 @@ void set_htotal(uint16_t htotal) {
 
   // hbst (memory) should always start before or exactly at hbst (display) for interlaced sources to look nice
   // .. at least in PAL modes. NTSC doesn't seem to be affected
-  uint16_t h_blank_start_position = (uint16_t)((htotal * (32.0f / 45.0f)) + 1) & 0xfffe;
+  uint16_t h_blank_start_position = (uint16_t)(((uint32_t) htotal * 32) / 45 + 1) & 0xfffe;
   uint16_t h_blank_stop_position =  0;  //(uint16_t)htotal;  // it's better to use 0 here, allows for easier masking
-  uint16_t h_sync_start_position =  (uint16_t)((htotal * (172.0f / 225.0f)) + 1) & 0xfffe;
-  uint16_t h_sync_stop_position =   (uint16_t)((htotal * (62.0f / 75.0f)) + 1) & 0xfffe;
+  uint16_t h_sync_start_position =  (uint16_t)(((uint32_t) htotal * 172) / 225 + 1) & 0xfffe;
+  uint16_t h_sync_stop_position =   (uint16_t)(((uint32_t) htotal * 62) / 75 + 1) & 0xfffe;
 
   // Memory fetch locations should somehow be calculated with settings for line length in IF and/or buffer sizes in S4 (Capture Buffer)
   // just use something that works for now
   uint16_t h_blank_memory_start_position = h_blank_start_position - 1;
-  uint16_t h_blank_memory_stop_position =  (uint16_t)htotal * (41.0f / 45.0f);
+  uint16_t h_blank_memory_stop_position =  (uint16_t)(((uint32_t) htotal * 41) / 45);
 
   writeOneByte(0xF0, 3);
 
@@ -1425,10 +1425,10 @@ void set_vtotal(uint16_t vtotal) {
   // VS start: 961 / 1000 = (961/1000)
   // VS stop : 964 / 1000 = (241/250)
 
-  uint16_t v_blank_start_position = vtotal * (24.0f / 25.0f);
+  uint16_t v_blank_start_position = ((uint32_t) vtotal * 24) / 25;
   uint16_t v_blank_stop_position = vtotal;
-  uint16_t v_sync_start_position = vtotal * (961.0f / 1000.0f);
-  uint16_t v_sync_stop_position = vtotal * (241.0f / 250.0f);
+  uint16_t v_sync_start_position = ((uint32_t) vtotal * 961) / 1000;
+  uint16_t v_sync_stop_position = ((uint32_t) vtotal * 241) / 250;
 
   // write vtotal
   writeOneByte(0xF0, 3);
@@ -1584,7 +1584,7 @@ void aquireSyncLock() {
   Serial.print(F(" Start HTotal: ")); Serial.println(htotal);
 
   // start looking at an htotal value at or slightly below anticipated target
-  htotal = ((float)(htotal) / (float)(outputLength)) * (float)(inputLength);
+  htotal = ((uint32_t)htotal * inputLength) / outputLength;
 
   uint8_t attempts = 40;
   while (attempts-- > 0) {
--- a/gbs-control.ino
+++ b/gbs-control.ino
@@ -116,23 +116,6 @@ void writeBytes(uint8_t slaveRegister, uint8_t* values, int numValues)
   writeBytes(GBS_ADDR, slaveRegister, values, numValues);
 }
 
-void writeProgramArray(const uint8_t* programArray)
-{
-  for (int y = 0; y < 6; y++)
-  {
-    writeOneByte(0xF0, (uint8_t)y );
-    for (int z = 0; z < 16; z++)
-    {
-      uint8_t bank[16];
-      for (int w = 0; w < 16; w++)
-      {
-        bank[w] = pgm_read_byte(programArray + (y * 256 + z * 16 + w));
-      }
-      writeBytes(z * 16, bank, 16);
-    }
-  }
-}
-
 void writeProgramArrayNew(const uint8_t* programArray)
 {
   int index = 0;
@@ -277,95 +260,6 @@ void writeProgramArrayNew(const uint8_t* programArray)
 // S1_00 - S1_2a "IF"
 // S3_00 - S3_74 "VDS"
 //
-
-// This is still sometimes useful:
-void writeProgramArraySection(const uint8_t* programArray, byte section, byte subsection = 0) {
-  // section 1: index = 48
-  uint8_t bank[16];
-  int index = 0;
-
-  if (section == 0) {
-    index = 0;
-    writeOneByte(0xF0, 0);
-    for (int j = 0; j <= 1; j++) { // 2 times
-      for (int x = 0; x <= 15; x++) {
-        bank[x] = pgm_read_byte(programArray + index);
-        index++;
-      }
-      writeBytes(0x40 + (j * 16), bank, 16);
-    }
-    for (int x = 0; x <= 15; x++) {
-      bank[x] = pgm_read_byte(programArray + index);
-      index++;
-    }
-    writeBytes(0x90, bank, 16);
-  }
-  if (section == 1) {
-    index = 48;
-    writeOneByte(0xF0, 1);
-    for (int j = 0; j <= 8; j++) { // 9 times
-      for (int x = 0; x <= 15; x++) {
-        bank[x] = pgm_read_byte(programArray + index);
-        index++;
-      }
-      writeBytes(j * 16, bank, 16);
-    }
-  }
-  if (section == 2) {
-    index = 192;
-    writeOneByte(0xF0, 2);
-    for (int j = 0; j <= 3; j++) { // 4 times
-      for (int x = 0; x <= 15; x++) {
-        bank[x] = pgm_read_byte(programArray + index);
-        index++;
-      }
-      writeBytes(j * 16, bank, 16);
-    }
-  }
-  if (section == 3) {
-    index = 256;
-    writeOneByte(0xF0, 3);
-    for (int j = 0; j <= 7; j++) { // 8 times
-      for (int x = 0; x <= 15; x++) {
-        bank[x] = pgm_read_byte(programArray + index);
-        index++;
-      }
-      writeBytes(j * 16, bank, 16);
-    }
-  }
-  if (section == 4) {
-    index = 384;
-    writeOneByte(0xF0, 4);
-    for (int j = 0; j <= 5; j++) { // 6 times
-      for (int x = 0; x <= 15; x++) {
-        bank[x] = pgm_read_byte(programArray + index);
-        index++;
-      }
-      writeBytes(j * 16, bank, 16);
-    }
-  }
-  if (section == 5) {
-    index = 480;
-    int j = 0;
-    if (subsection == 1) {
-      index = 512;
-      j = 2;
-    }
-    writeOneByte(0xF0, 5);
-    for (; j <= 6; j++) {
-      for (int x = 0; x <= 15; x++) {
-        bank[x] = pgm_read_byte(programArray + index);
-        if (index == 482) { // s5_02 bit 6+7 = input selector (only 6 is relevant)
-          if (rto->inputIsYpBpR)bitClear(bank[x], 6);
-          else bitSet(bank[x], 6);
-        }
-        index++;
-      }
-      writeBytes(j * 16, bank, 16);
-    }
-    resetPLL(); resetPLLAD();// only for section 5
-  }
-}
 
 void fuzzySPWrite() {
   writeOneByte(0xF0, 5);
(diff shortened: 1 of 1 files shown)


bkoropoff:
Upped the maximum correction factor a bit and now it works with my NTSC SNES

ramapcsx2:
Awesome!
I'm currently quite ill and don't feel much like coding work but once I get better, I'll try this out.

I recently discovered that on the left side of the scaler chip, there is a single pin broken out to the original controller.
That pin belongs to a digital video input / output, so that's kind of weird.
Well, it turns out that this pin can be configured to output debug information such as the decoded input hsync and vsync pulses.
It's a little hard to find a good solder point for it, but I could make good use of this in my sync lock.
Using this, I wouldn't have to switch the output off like I currently do, and just measure the incoming timings continuously. Also any sync degradation could be seen in that signal, enabling some software countermeasures.

ramapcsx2:
Tons of changed code but once thing stood out:
You seem to be using set_vtotal() for your adjustments? 
The h_total, v_total functions are meant for quick dialing in some sane numbers when creating new presets / using different pll clocks.
As such, they change more parameters than required. A single call to set_vtotal() on a normal preset would change the picture geometry quite drastically.
But this is easy to work around: Simply create a smaller version of set_vtotal() that works for live updates.

The bigger issue is that on no display that I've tried, was it possible to change the amount of active vlines without causing a picture dropout.
Maybe your display is the exception? Otherwise you'd surely have noticed it.
The problem is less pronounced on the horizontal timings, where it's usually possible to add or subtract ~3 pixels until the display resyncs.

bkoropoff:
Yeah, I used set_vtotal because it was quick and easy, and because I'm running low on available prog mem.  The preset format probably needs to be changed to specify small deltas on top of a base configuration in order to reclaim more memory.

I found on my display that changing the length of the vsync pulse causes dropout, and changing the length of the back porch causes the vertical position of the picture to shift (which makes sense).  Changing the front porch is well-tolerated as long as it doesn't change too quickly, which is part of why I added some damping to the correction code.  Testing on other displays is required to see how they cope with it.

ramapcsx2:
Hm, I don't yet understand how exactly this works but I'm sure I'll see when I try it.
May I ask you to change this PR so it uses it's own "set_vtotal()" with just the stuff that it needs?
All the rest is fine by me, if it's necessary on Arduino. 

(Mind though that I found that the ATmega328 flash limit is far too easily maxed out with this project, something I didn't realize at first. Considering also that people need some form of cheap and readily available control method, I recommend you switch over to ESP8266 ;) )

bkoropoff:
@ramapcsx2 I confirmed I got the desired behavior by hooking up my scope to csync input and vsync output.  Without locking, you can observe the vertical sync pulses drifting with respect to each other and eventually crossing, which results in tearing.  With locking, the drift reverses direction every time the pulses get too close or too far away, which keeps the output vsync safely trailing the input vsync by part of a frame.

ramapcsx2:
Ah, you probably have a digital scope?
I have a lot of trouble working with the rather long vertical intervals on my 1983 Iwatsu, so while I can probably set this up with some work, it doesn't come easily ;p

But thanks for that extra explanation, I think I get the idea now. Now I just wonder why I don't notice this as a problem here. Must be some other variable in the long chain of analog video stuff. That, or I just don't notice it ;p

(I am in the process of hooking up my ESP board to try this. But thinking about it, the ADC on these boards works differently to Arduino and many people say it's worse. Maybe better I dig out the Uno?)

ramapcsx2:
Okay, I pushed this onto my Uno. Connected csync to A0 and got this in the log:

samples: 55473
csync min level: 22
csync max level: 585
csync true level: 303
csync min pulse (us): 20
csync max pulse (us): 204
csync pulse threshold (us): 112

My PC TFT does not reliably lock onto the signal. The best it can do is stay stable for about half a second, before going into constant sync loss / recover cycles.

Any values to tweak?

bkoropoff:
The vtotal is supposed to change.  You can uncomment the serial prints in `doVsyncPhaseLock()` to see what phase offsets it measures and corrections it uses.

I'll test on some different displays later today and see if they show the problem you're seeing.  I might have been lucky with my TV being unusually forgiving about vlines.  It should be possible to rework the technique to change htotal instead.  The important thing is that the total frame/field time can be constantly varied in a negative feedback loop keep the vsync phases from drifting.

ramapcsx2:
Okay, I will wait for a htotal version of this, in case you want to try it.
When you do design such a thing, please be aware that even that might not be stable enough.
Messing with the timing intervals is poorly received by almost all devices.

But we can still try things. There is an entirely different "retiming" input sync processing mode that I recently explored, but turned off again. It allows reshaping the regenerated sync pulses, for example, and it might open some other way to deal with the drift.
Should you want to test that mode, you'll need your scope ready, and a wire on that debug pin I mentioned earlier.
(I'll do a photo then)

ramapcsx2:
Oh, and I just remembered:
There is a promising option on the VDS as well, called something like "different frame size".
It allows you to set up additional canvases / timings and then (I think!) it works together with the frame lock mode.
You could, for example, tell it to use frame format 1 for 5 consecutive frames, then frame format 2 for the 6th and back to start.
This is probably more compatible than sporadically altering the timings in real time, as there's at least a nice pattern to it.
But I don't fully understand it yet, so I probably haven't seen all that this can do.

I did get regular frame lock mode to stay synced for minutes before though, so there's that :)

bkoropoff:
@ramapcsx2 Yeah, I remember noticing that in the register documentation and thinking that it may have something to do with how frame lock is supposed to work.

I tested htotal- and vtotal-based locking with someone else's display, and both caused sync drop-outs.  I have a third display that I plan to test tonight to get an extra data point.  One thing I'd like to test is whether waiting until vsync/vblank/vactive to make an adjustment affects the ability of the display to cope.  However, I suspect that the 5725 is doing some latching of the registers, so I may not have that much control over it.

bkoropoff:
Thanks.  I found that my Samsung LCD monitor wouldn't lose sync, but it did show display artifacts when the output timings changed.  I plan to do some more experiments to see if that can be eliminated.  If I can get this working on at least a couple major brands of display then it might be worthwhile as a non-default option.  I suspect this would also be better tolerated by actual CRT displays, but I don't have one handy to test with.

ramapcsx2:
Sounds like a plan :)
My display seems to accept HTotal changes of +/- 5, which is pretty big already. It's enough to correct a PSX switching from regular 240p to interlaced (where the timing is slightly different).
Somehow this only started working once I used the new debug pin. No idea what's up with that.

ramapcsx2:
I found something that you can probably use!
Try setting this:
s3_21 + s3_20 to "1000" (s3s21s03 s3s20se8)
s3_23 + s3_22 to "1000" (s3s23s03 s3s22se8)
Then enable "different frame size" and "repeat nr."to "2":
t3t1ft0 t3t1ft4

Now one more thing:
Set s3_1b to "2": s3s1bs02

Now you can modify s3_22 and switch between "e8" and some other, nearby value. Only "ea" works here but my display is picky with vertical timings.
But even on my display, I can toggle the s3_22 value at any time, and doing so changes the direction of the screen tear. Ie: It enables your fix but is more compatible :)

bkoropoff:
Thanks, I have a few other projects I'm juggling, but I'll take a look this week.

ramapcsx2:
I have a hard time figuring out your pacing logic. I'd implement this if I knew what you do to determine when it has to speed up or slow down.
With the debug pin, you can put the incoming vsync signal on there (which it is by default, but other signals are available as well, such as HSync).

bkoropoff:
I can get the vsync phases to lock by adjusting s3_22, but any correction besides 0 or 1 causes extreme flickering, as if every other frame were black.  For reference, syncing with my SNES tends to alternate between correction values of -4 and -5.

ramapcsx2:
Okay, maybe you've already checked with the register definitions but it looks like I gave you instructions for 2 different variable frame sizes. I guess only one is necessary, minimizing potential issues.
The variable frame setting that worked was "e8" and "ea". Anything else would drop out.
You don't also have frame lock enabled by chance, do you? It doesn't have to be on to work.

ramapcsx2:
By the way, I can push a commit with just the floating point math stuff, if that helps you with your branch.

bkoropoff:
Yeah, setting a frame size of +4 over VSYNC_RST greatly upsets my television, to the point that it stops refreshing the LCD entirely.  This is without any dynamic correction at all.

bkoropoff:
@ramapcsx2 It can tolerate anything between 0xe8 and 0xea, but that range of values doesn't allow aligning vsync, at least with my SNES.

ramapcsx2:
Ah okay, then maybe you can change your approach to correction:
On my SNES, "e8" makes the line go down and "ea" makes it go up. So in theory, it now becomes a problem of timing that switch so the bar always stays off screen.
It should work the same for you, but only if the regular frame lock put HTotal within the target range first.

bkoropoff:
The thought occurred to me, yeah.  The initial video timings need to be adjusted so the output frame time is slightly less than the input, after which small positive adjustments to the frame size should be able to hold them in sync.

bkoropoff:
I'll also try wiring up the debug pin at some point.  If it works for getting vsync then the ADC calibration stuff can be thrown out.

ramapcsx2:
It does that :)
It appears to be the output of the SOG separator, so it's only as good as that part is working.
Basically, if you use Composite video for sync and do a lot of white > black transitions, it will be unstable and jitter.
But at that point, we have other issues than caring for frame rate compensation. The chip won't even sync up properly ;p

bkoropoff:
OK, I pushed a new version of my branch which is a proof-of-concept of melding the two approaches.  It's not ready to merge yet since it has a bunch of debug output turned on.

When initializing sync locking, it attempts to find the largest value of htotal for which the output frame time is smaller than the input.  This means that without any correction, the (output - input) phase offset will creep downward, and increasing the frame size by a small amount (2, in this case) should cause it to creep upward.  The phase locking logic is a bit simpler since we only ever pick between two correction values based on what the phase offset is.  It tries to keep the output trailing the input by 90 degrees, which in my testing seems to keep the tear off screen.

ramapcsx2:
Great!
I just pulled the branch and put this onto my Arduino.
It looks like it hangs after the last line in the log:
post preset done
samples: 5523
csync min level: 0
csync max level: 380
csync true level: 190
csync min pulse (us): 20
csync max pulse (us): 176
csync pulse threshold (us): 98
Base htotal: 1803
Test htotal: 1803

ramapcsx2:
Yep, this is working :)
My display is fine with the adjustments and the tear line if kept off screen.

It would be great if you used my frame lock routine with the debug pin (replacing your findBestHTotal()).
It may already find the htotal value you require. If not, it will be simple to nudge it towards that value.

Next, this should all work without having csync, just using the debug pin.
If you require hsync on that pin, set S0_4d to 0x2c. If you want vsync on that pin, set S0_4d to 0x2a.
0x2a mode can be further configured via S5_63 but that is already set to output vsync.

Great work :)

bkoropoff:
Pushed some more changes that makes it lock in on the correct htotal more quickly.  It now also works on my Genesis.  The code is a bit ugly because of the ADC stuff.  I can switch to using the debug pin, but we'll have to make it mandatory rather than an option in order to remove the ADC pin.

Because the frame size correction is now so small, I found that there is precisely one value for htotal where the phase offset shrinks at `frame size = vtotal` and grows at `frame size = vtotal + 2`.  If any of the heuristics in the original sync setup code nudge htotal off of that value, sync locking won't work.  Because of that it was easier to start from scratch.  We should look at preserving some of the blanking changes you made, however.

bkoropoff:
Alright, I got it working using the debug pin instead.  It's easier to solder to anyway!  The amount of jitter in reading the vsync periods is concerning, but I seem to reliably get the right htotal.

ramapcsx2:
If you get high jitter (that would be visible on the scope), then that's a sync processor issue and can be fixed. I just don't know this unit well enough to do it for all sources of sync and all consoles.
It could also be a hardware issue. My 1Chip SNES produces jitter only for the first few minutes of operation. That may have to do with the resistor divider solution I did for csync on that console.

When testing this code yesterday, I found that the htotal solution looked a lot like what I started with.
Mine got very good over time though, determining the correct and same htotal 90% of the time, with the rest being one over target.
I never had any better results, likely due to jitter in the measurements themselves (the results improved when introducing nointerrupts() but those vsync periods are rather long anyway).

I'll check out your changes :)

ramapcsx2:
Regarding the debug pin vs csync from a sync stripper or console:
It's hard to tell what hardware people have. Some may use YUV inputs a lot. If there's no csync available, the new method won't work.
But we know that every GBS board has a debug pin that works with all kinds of inputs that the GBS itself can lock on to.

An idea over at circuit-board.de is to build clip on adapters using a PLCC44 socket, breadboard and MCU.
This makes all signals available without requiring any soldering. Perfect for end users!




An extra benefit is that with this pin connected, we can analyse the extracted sync quality and try to fix problems.
For that part, it would probably be best to regularly check this pin, so maybe in an interrupt routine.
But that's for later ;)

ramapcsx2:
The best htotal algorithm is excellent. I just hope that measurement and / or signal jitter won't put it off course. But if they do, it should be possible to fix it later.

I found a corner case problem with my PSX in PAL mode. The algorithm can fail with an off by one error if the starting htotal == best htotal. A targeted fix for it is something like this:

```
static uint16_t findBestHTotal(void) {
  uint16_t htotal = readHTotal();
  uint16_t start_htotal = htotal;
  uint16_t min = htotal - htotal / 100;
  uint16_t max = htotal + htotal / 100;

  Serial.print("Base htotal: "); Serial.println(htotal);

  while (min < max) {
    htotal = (max + min) / 2;
    Serial.print("Test htotal: "); Serial.print(htotal);
    writeHTotal(htotal);
    delay(100);
    if (outPeriodIsLarger()) {
      Serial.print(" (too high): ");
      max = htotal - 1;
      Serial.println(max);
    } else {
      Serial.print(" (too low): ");
      min = htotal + 1;
      Serial.println(min);
    }
  }
  
  Serial.print("Best htotal: "); Serial.println(min);
  if (min == (start_htotal + 1) ) {
    Serial.println("correcting for overshoot");
    min-=1;
  }
  
  return min;
}
```

bkoropoff:
Can you post the serial output when it overshoots by 1?  I don't have a PAL or PAL-modded system to test, but I can try to tweak the logic.  One thing I considered doing is adding a fuzz value so it will err on the side of picking a smaller htotal if the value it tests is extremely close.  The fuzz should be equal to approximately the half the duration of a scanline from the input.  Is there an easy way to obtain this value?

ramapcsx2:
Here are my changes, fixing both issues mentioned. This now works well in all 4 modes my PSX can produce (PAL regular, interlaced & NTSC regular, interlaced).

https://github.com/ramapcsx2/gbs-control/commit/7abd30c09d4cc7d4fe8e5d8c0e53abf93dd5c66a

bkoropoff:
OK, I pushed two changes that should fix the off-by-one and interlacing issues.  The off-by-one was a simple tweak to the binary search logic.  I can confirm it works here when I force the first guess to be the correct htotal.  The interlacing fix just samples two consecutive input frames and divides by two as you said.

ramapcsx2:
Okay, I'll test it as well. If it works, this PR needs a couple more safety checks and then I'll merge it.
I just want to make sure it doesn't blow up expensive displays or something ;)

bkoropoff:
This still doesn't work on my Samsung monitor, and it doesn't work through my VGA-to-HDMI converter on either the monitor or my TV.  The active phase locking part of this probably needs to be behind an option if it gets merged.

ramapcsx2:
Bummer. Thanks for letting me know.

ramapcsx2:
Okay, I hope the merge went well.

Thanks a ton for your nice work! :)

PR #11, opened 2018-04-09
Work-in-progress active sync locking tweaks

bkoropoff:
It turns out that shifting the vsync pulse when increasing vtotal prevents the effective vblank region from moving around.  It was only stable for me, however, after I had adjusted the baseline vblank start/stop to crop the overscan area and center the rest of the screen.  I think the larger front/back porches were necessary to give some "wiggle room" for the vsync pulse to move.  Without doing this, my monitor would shift the picture up/down or lose sync entirely.

With this configuration, my screen is completely stable, centered, and clean on an NTSC Genesis.

sonic

I'm not sure I have a complete grasp of how all the registers interact with each other and the display.  Please play around with this and let me know what you find.  I still need to try this on my TV and with my NTSC SNES.


bkoropoff:
After reverting b962fd9edd8bd to get my SNES syncing again, this seems to be stable with clean borders and a correctly-centered image.  My TV is also fine with it.  I can test my other Samsung monitor for a 3rd data point, but it's never worked with any of the previous active sync methods so I doubt it's going to start working now.

rockman

bkoropoff:
Reverting b962fd9 also causes the active region of the screen to be misaligned when going back to my Genesis.  It looks like input formatter registers also have to be carefully calibrated to get correct screen geometry.

I've found that my monitor likes to guess a new vertical position for the screen every time the video output is cycled, but I'm not sure there's much that can be done about that.  The active region of the signal always seems to encompass the correct portion of the Genesis output (e.g. it excludes the blue borders in Sonic games).

My newer Samsung monitor continues to freak out when vtotal is changed at all.

Maybe I should just buy an OSSC.

ramapcsx2:
I'll try this branch, see how it fares here.
For the image positioning, this is an issue of blanking and cropping, as well as IF tweaks.
For example, you could extend vertical memory blanking again. I tried fixing issues I saw by making it very small (just a single line), maybe your setup requires it larger.

bkoropoff:
Yeah, I'm still having htotal problems here as well.  I'm going to try a different approach.

Yes, the basic difference is that the vsync pulse moves forward with the vtotal.  With this arrangement, the two extra scanlines occur in the vertical front porch.  The interval between the end of the vsync pulse and the start of active video for the next frame remains the same as long as vblank-stop occurs after the vertical counter has wrapped around to 0, which is why I needed to change the preset to get stable results.  If vblank-stop occurs before wrap-around (e.g. it equals 999), it would need to move as well.  I'm not sure if I should change the code to do this, or if we should just make sure all the presets have vblank-stops that occur after wrapping.

bkoropoff:
OK, I seem to get good results when averaging two samples, and I've also added a check that we get the same best htotal 4 times in a row before accepting it.  This seems to terminate quickly, and at any rate we aren't actually changing the htotal in the loop, so it's less important that it be perfectly fast.

bkoropoff:
The frequency of bad htotal results before it gets 4 good ones in a row is concerning.  It might be better to produce a fixed number of candidates (say, 10) and then pick the most frequent, but I'm trying to go easy on the memory and code footprint.

ramapcsx2:
When starting this project, it looked like Arduinos would be perfect for it.
As it turns out, all the things that need to be done and options to support do require a bigger MCU.
I recommend everyone switch to ESP8266.
These boards are often even cheaper than Arduino clones, and they offer a lot more.
The web ui for example is really helpful when tuning presets.
I wouldn't be surprised if the faster I/O speed and much faster core would help with measurement jitter as well. "SyncHtotalStable" is simply not required here (but that's only true when sampling 2 fields of the input, due to the possibility of interlaced sources).

bkoropoff:
I got fed up with manual bit-twiddling everywhere, so I added some template code to access logical TV5725 registers, even when they have bit-level offsets.  Long story short, you can define a register in `tv5725.h` like this:

```C++
  // Segment 3, byte offset 2, bit offset 4, 11 bits
  typedef UReg<0x03, 0x02, 4, 11> VDS_VSYNC_RST;
```
You can then use it like this:

```C++
uint16_t vtotal = GBS::VDS_VSYNC_RST::read();
GBS::VDS_VSYNC_RST::write(vtotal + 1);
```

bkoropoff:
Hmm, this was working for me but now there are problems of some sort.

bkoropoff:
OK, fixed

ramapcsx2:
Yeah, I like those.
Just hope that huge register template is fine ;p

ramapcsx2:
Sorry for creating a conflict here. Didn't think it would.

This PR could be separated.
Dumping registers with comments and debug templates are great and could be merged.
The more reliable best htotal calculation commit seems fine as well.
Those things could go in right now.

The changed NTSC template is customized for SNES and Genesis. I'd like to keep the regular presets multi purpose (so including later, often interlaced sources).
We could have new presets for specific consoles that people can select from.

The new template based controls seem to be buggy sometimes and some things don't work at all.
Changing the scaling factor ('h', 'z', '4' and '5') is broken.
New sync lock seems to work at first, but after changing presets (maybe a few times), it doesn't seem to keep the lock.
This might be related to the templated controls or even me using an ESP32 at the moment.
I'll investigate but it would be great if you prepared a PR with just the stuff listed, for merging as is :)

bkoropoff:
Looks like scaling controls don't work because it was implicitly depending on readFromRegister to change the current segment.  Easy to fix.

Changing the preset probably needs to clear syncLastCorrection or it will think one is already applied and get mixed up.

bkoropoff:
The NTSC template was calibrated against Genesis, but it probably reflects what portions of the screen developers of the era expected to be in the overscan area.  It's probably fine as a starting point for making custom adjustments which can be saved in some way.

bkoropoff:
Added some more changes.  Removing the vsync/debug connection tests should now be safe because `initSyncLock` will simply time out and give up if one of the connections is missing.

ramapcsx2:
> Looks like scaling controls don't work because it was implicitly depending on readFromRegister to change the current segment. Easy to fix.

Yeah, readFromRegister with segment changes was a bad idea. I'll replace them all if you haven't done so yet.

> it probably reflects what portions of the screen developers of the era expected to be in the overscan area

It's true that developers expected quite a bit of overscan. I would favour including cropping for that but there are some issues still. One problem is that the GBS has a concept of vertical active video start, which I think they expect you to detect from the signal. I don't know how to do this yet, so many consoles are vertically shifted from each other.
The most obvious example can be seen with 1Chip SNES's that seem to put the first visible scanlines earlier than other consoles.

To make this all work anyway, I regularly calibrate against a known reference, which is the 240p test suite grids on a PSX.

Now, when I merge your presets with the cropping, it would most likely crop my test pictures. 
I'd hate to loose those as a reference, because the PSX is mostly standards conform with respect to other (especially later) consoles, and I can switch it easily to PAL and NTSC.

I'll try your preset with the PSX later and if the grid is mostly there, I'll go with it.
Hopefully we can offer OSSC like optional masking some day that just works ;)

>  Removing the vsync/debug connection tests should now be safe because initSyncLock will simply time out and give up if one of the connections is missing.

Sounds good!

> The mode detect reset patch works for me, but I'm not sure if it's the right approach.

I'll try it out. There are many test buses that I just discovered but so far I haven't noticed a definitive "source is on" signal. It must be there though.

ramapcsx2:
The cropping doesn't seem too bad :)

ramapcsx2:
The template code throws quite a few warnings. Are those reasonably fixable?
Try setting compiler warnings to "all" in the Arduino IDE preferences to see them. 
The ESP32 compiler doesn't even allow compiling due to "some warnings being treated as errors" :p 

The mode detect reset is working, but it would be best to have the information just via register reads (for people with no debug wires). I've already found good indicators on the SP test bus, but nothing conclusive yet.

Removal of the debug / vsync tests is okay with the graceful exit.

The cropping does affect the PSX grid test but oh well, it can be undone if required.

One more thing to try about measurement jitter:
digitalRead() is very slow on Arduino (I think it's no problem on the faster MCUs).
It may be too slow for accurate readings. You could try direct pin manipulation like so:
https://github.com/ramapcsx2/PsNee/blob/master/PsNee.ino#L313
It makes that code less portable, since each MCU has their own way of doing this, but it's fine in an #ifdef :)

bkoropoff:
The compiler warnings were harmless (warning about truncation I wanted on purpose, and a false positive about shifting by a negative amount that can't actually happen).

ramapcsx2:
Great, I'll merge the pr after work then!

PR #13, opened 2018-04-10
Remove vsync/debug connection tests

bkoropoff:
These connections are basically mandatory, and at any rate the test only works if the console is turned on before the controller so there is a video signal during `setup`!

### DIFF
--- a/gbs-control.ino
+++ b/gbs-control.ino
@@ -2074,29 +2074,6 @@ void setup() {
     delay(1000); // at least 750ms required to become stable
   }
 
-  // check whether vsync in and debug in are connected and working
-  boolean VSYNCINconnected = false;
-  boolean DEBUGINconnected = false;
-  if (pulseIn(vsyncInPin, HIGH, 100000) != 0) {
-    VSYNCINconnected = true;
-    Serial.println(F("VSYNC-IN connected :)"));
-  }
-  else {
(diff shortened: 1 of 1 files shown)


bkoropoff:
I've been experimenting with monitoring the input vsync through the debug pin to decide when to reset the mode detector, which has fixed problems with it getting stuck when turning off the console.  I think the debug pin is going to be indispensable sooner or later.

Soldering to it is isn't *too* bad.  Soldering to one of the TV5725 pins looks like a real pain, though.

bkoropoff:
Let me know how the sockets work.  If they're a good solution I might take a crack at putting together a PCB layout.

The current mode detection basically always fails to detect my console turning off and requires the controller to be reset.  If I reset mode detection whenever phase measurement times out (e.g. they are no discernible input pulses), it gets unwedged.

ramapcsx2:
I kind of bought the wrong sockets but still managed an adapter proof of concept.
The problem with the sockets is that their pin pitch is very narrow and they're all in a single row.
This requires some tricks when soldering, especially on the SCA and SCL pins, as they're next to each other.

Pics are here :)
https://circuit-board.de/forum/index.php/Thread/15601-GBS-8220-Custom-Firmware-in-Arbeit/?postID=704650#post704650

PR #15, opened 2018-04-13
Preliminary OSD support

bkoropoff:
This allows changing screen alignment and size through the built-in OSD.  I'm using the buttons on the GBS board itself to control it, but it would be easy enough to hook up an IR receiver or similar.

### DIFF
--- a/gbs-control.ino
+++ b/gbs-control.ino
@@ -62,6 +62,19 @@ extern "C" {
 
 typedef TV5725<GBS_ADDR> GBS;
 
+enum class MenuInput {
+  UP,
+  DOWN,
+  FORWARD,
+  BACK
+};
+
+enum class MenuState {
+  OFF,
+  MAIN,
+  ADJUST
+};
+
 //
 // Sync locking tunables/magic numbers
 //
@@ -1574,6 +1587,8 @@ void resetRunTimeVariables() {
 void doPostPresetLoadSteps() {
   // Any sync correction we were applying is gone
   syncLastCorrection = 0;
+  // Any menu corrections are gone
+  menuInit();
 
   if (rto->inputIsYpBpR == true) {
     Serial.print("(YUV)");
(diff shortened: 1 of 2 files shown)


ramapcsx2:
Do tell when you want this merged. 

I don't have the setup to test this, but I heard others mention IR support at least.
Wiring up all 4 buttons seems like a hassle now, especially with the web ui available.
They will be easy to support with the plug and hack adapters though :)

ramapcsx2:
Hah, I found a couple unused command handler keys and can test this now. 
Awkward, but it works :)

The menu works very well already. I can see how this is useful with the buttons hooked up.
I guess I'll spend some time looking for free I/O pins on ESP32 and ESP8266 now.

One thing I noticed:
The button handler runs very frequently. Every time it runs, it first debounces the buttons.
This won't take long, of course, but if the loop finishes quickly / instantly, it might add up.

If it does add up, then the button check could run from an interrupt handler.
In that case, it only ever gets to work once a button press has been detected.

bkoropoff:
I disabled `HAVE_BUTTONS` by default since pins left floating will cause phantom inputs.  That means this can be safely merged.

ramapcsx2:
Yep!

gingerbeardman:
Hi @bkoropoff @ramapcsx2 how do I go about using this? Wire up the buttons? If so, how? Thanks!

ramapcsx2:
Where do you need help? :)

Pretty rough overview:
The Arduino code sets up bitfields with all the option bits, then sends them on each websocket update.
The Javascript should parse the bitfields and light up any set buttons as required.

ramapcsx2:
Sorry if you got an Email with an earlier reply. I thought this was about the new web ui :)

For the OSD stuff, yes, you need to wire each button to a free GPIO (input) on the ESP8266.
It will be hard to find 4 free pins..

gingerbeardman:
Ah, OK. 

I'd like to contribute code and figured OSD with GBS hardware buttons would be a good pairing with WebUI. We'll see.

PR #17, opened 2018-04-14
Add and use interface to combine register reads/writes

bkoropoff:
This allows multiple registers in the same segment to be updated with a single I2C read and write cycle.

### DIFF
--- a/gbs-control.ino
+++ b/gbs-control.ino
@@ -1490,24 +1490,21 @@ bool vsyncPeriodAndPhase(uint32_t* periodInput, uint32_t* periodOutput, int32_t*
 }
 
 void adjustFrameSize(int16_t delta) {
-  uint16_t vtotal, vsst, vssp;
-
-  vtotal = GBS::VDS_VSYNC_RST::read() + delta;
-  vsst = GBS::VDS_VS_ST::read() + delta;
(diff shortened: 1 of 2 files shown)


ramapcsx2:
Holy.. I've got to learn that stuff before merging, lol. ><

bkoropoff:
Added some comments to the code for clarity

ramapcsx2:
```
typedef GBS::Tie<GBS::VDS_VSYNC_RST, GBS::VDS_VS_ST, GBS::VDS_VS_SP> Regs;
auto plusDelta = [=](uint16_t x) { return x + delta; };
Regs::modify(plusDelta, plusDelta, plusDelta);
```
I'm sorry but is this really necessary?
It's nice if this saves some I2C traffic but is it worth the obfuscation?
Once merged, will there be more of this?

If there's any kind of bug in those new templates, I certainly will never find it.

bkoropoff:
I can change it to a read followed by a write.  That retains the main
advantage, which is that the video timings change as close to
simultaneously as possible.

ramapcsx2:
That would be easier to work with, yeah.
I didn't realize that this was to tighten video timing changes.
That's probably a good thing to do.

bkoropoff:
Changed

PR #18, opened 2018-04-14
Add fast digital pin read function

bkoropoff:
The hope is that the call gets optimized to a direct memory read by including the pin lookup tables directly.  Let me know if this works on the other boards you have.

### DIFF
--- a/fastpin.h
+++ b/fastpin.h
@@ -1,0 +1,11 @@
+#include <pins_arduino.h>
+
+static inline uint8_t volatile* portToInput(uint8_t port) {
+    return reinterpret_cast<uint8_t volatile*>(port_to_input_PGM[port]);
+}
+
+template <uint8_t Pin>
+static inline bool fastRead() {
+    return (*portToInput(digital_pin_to_port_PGM[Pin]) &
+            digital_pin_to_bit_mask_PGM[Pin]);
+}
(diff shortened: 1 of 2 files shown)


ramapcsx2:
The other boards don't have convenience accesses such as "port_to_input_PGM[]" in their pins_arduino.h.

This could get a little tricky, too. 
ESP8266 is pretty normal still, with just a couple variants in the pinouts, that can all be grabbed from their  pins_arduino.h (people have to select their board from the tools menu, which then includes the correct version).
ESP32 is weirder though, in that most of the pins can be configured to be I/O.
I hope the Arduino device selection makes this possible as well though. If it does, it will be the same as ESP8266.

People have already figured out fast methods by the way. Here are some methods:
http://skpang.co.uk/blog/archives/323
https://www.codeproject.com/Articles/732646/Fast-digital-I-O-for-Arduino

( Wow, a regular digitalRead() takes 4us >< )

ramapcsx2:
I'm just working on fast digitalRead for ESP8266.
It actually makes a difference!
Regular digitalRead takes 77 cycles, direct register read takes 21 cycles.
This translates to a vsync sampling swing of 4 .. 8 milliseconds for regular, and 5 .. 7 ms for the fast version.

On Arduino, I'm sure the effect will be much greater.

PR #22, opened 2018-04-15
Fix fast digitalRead on Arduino

### LINKED ISSUE #21: Arduino replacement digitalRead() is not compatible with all boards
It does not work on my Pro Micro.  Only the values in pins_arduino.h are authoritative.

bkoropoff:
Closes #21


ramapcsx2:
Cheers :)

PR #24, opened 2018-04-15
Hoist sync locking code into its own class

bkoropoff:
This is mostly for improving code organization.  The `.ino` file is getting pretty large as new features are added.  Let me know what you think.


ramapcsx2:
Agreed, the sketch has become hard to navigate and this pr is a good effort to mitigate the problem.
There's just one nitpick I have: The class name should be more specific. "SyncManager" could be the SyncProcessor, for example.

How about "FrameLockManager" or "FrameTimeLockManager" or somesuch?
That's what the code does anyway: find out the best htotal to get close to frame time lock, then keep that lock continuously.

But otherwise, this is an improvement!

bkoropoff:
Renamed

PR #27, opened 2018-04-30
Don't use bool for 1-bit registers

bkoropoff:
This results in undefined behavior when attempting bit twiddling with the bool type.


ramapcsx2:
Nice, thanks for taking care of it :)

PR #182, opened 2021-01-29
Edtv passthrough

goutsune:


goutsune:
a...

gingerbeardman:
Please consider merging this @ramapcsx2

ramapcsx2:
Ah, I looked at this and there were problems. A lot of talk required here ><

goutsune:
> Ah, I looked at this and there were problems. A lot of talk required here ><

I'm all ears, I have been using gbs-control like this for about an year with PS2 and Wii, but perhaps I forgot to change some pre/post preset flags?

goutsune:
I understand that code is messy and mine no better, so I think it's ok to let people test it and decide :D 
I have been using GBS like this for a few years (+ I think I reduced some of input signal test delays to make CRT go to standby faster that in 5 or so minutes).

On a side note, I have been pondering around possibilities of true dynamic preset genetaion based on analyzed input signal — I noticed there are quite some test registers on GBS and I think it is possible to exploit clockgen remaining channels to try and catch pixel clock, kinda how LCD monitors do that when auto-matching screen.

ramapcsx2:
I think I've probed the registers thoroughly, but it remains truly black box reversing. If you find anything new that can be used, great, but I also already tried hard :)

PR #203, opened 2021-03-09
Fixes wifi input password don't accepting input

xperiments:
Fixes wifi input password don't accepting input


ramapcsx2:
Cheers :)

PR #205, opened 2021-03-10
Fix general input not responding to keyboard events

xperiments:
Fix general input not responding to keyboard events.

This is a more general fix than the previous applied.
Now all remaining inputs will be accesible.


ramapcsx2:
And done :)

PR #288, opened 2021-10-25
Fix compile error with more recent ESP8266 toolchains

M-Reimer:
With latest toolchains compiling fails in "framesync.h" at a position where the absolute value of the difference is calculated.

The problem with this is that both values, that are subtracted, are unsigned values. I don't know if older compilers had some workaround in place here, but this is clearly wrong. If you subtract a bigger unsigned integer from a smaller unsigned value, then this does not get negative but it will overflow and the result is clearly not what is expected here.

I've replaced this line with something that at least I think tells clearly what is meant here: "Subtract the smaller of the two variables from the bigger of the two variables".

### DIFF
--- a/framesync.h
+++ b/framesync.h
@@ -145,7 +145,7 @@ private:
     if (inPeriod == 0 || outPeriod == 0) { return false; } // safety
 
     // allow ~4 negative (inPeriod is < outPeriod) clock cycles jitter 
-    if (abs(inPeriod - outPeriod) <= 4) {
+    if (max(inPeriod, outPeriod) - min(inPeriod, outPeriod) <= 4) {
       /*if (inPeriod >= outPeriod) {
         Serial.print("inPeriod >= out: ");
         Serial.println(inPeriod - outPeriod);
--- a/framesync.h
+++ b/framesync.h
@@ -145,7 +145,7 @@ private:
     if (inPeriod == 0 || outPeriod == 0) { return false; } // safety
 
     // allow ~4 negative (inPeriod is < outPeriod) clock cycles jitter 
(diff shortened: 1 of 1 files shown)


M-Reimer:
Can you please have a look at my latest commit? This should be about as fast as we can go.

ramapcsx2:
Hey,
I just got back from a trip. I'll check it out soon!

LawnMo:
Wouldn't it be easier and preserve what was initially meant by kanging ABS() from : https://github.com/MarlinFirmware/Marlin/commit/99ecdf59af907ebb8d2d847863614094bb576e3f#diff-a426db550c957a2391fe722b87be2b9d666b39aad58d382be93670b0a25167e8R197 ?  
Haven't tried the actual change, but at least it compiles with latest toolchains.

M-Reimer:
> Wouldn't it be easier and preserve what was initially meant by kanging ABS() from [...]? Haven't tried the actual change, but at least it compiles with latest toolchains.

The only reason why this works is that the compiler does not know "ABS()" so the error check will not trigger.

The macro, used by the Marlin firmware, also does not "magically allow" to subtract a bigger from a smaller unsigned integer and get a negative value. C(++) just doesn't work this way. If you subtract unsigned integers, you get an unsigned integer. If the value, you subtract from, is smaller then it overflows. The "ABS()" macro is just a faster (probably not even in all situations and heavily dependent on compiler version) of the original "abs()".

ramapcsx2:
Sorry again for the delay. Your change is good, of course :)

I have to say that this code is ultimately relying on the speed of the ISR, how quickly the hardware reacts to a pulse and how quickly then the ISR code gets to execute.
It is probably tuned for the ESP + software as it was back then, which may or may not have changed by now.
All this is to say, it should be checked whether it still does the job :)

PR #334, opened 2022-06-07
Feature/issue 237 disable ext clock gen

fabbrimichele:
The purpose of this pull request is to add an option in the UI to disable the external clock generator.
This makes the option _Force PAL 50Hz to 60Hz_ work that otherwise is ignored when using the external clock generator.


ramapcsx2:
Thank you!
I'll review asap :)

ramapcsx2:
I'm so sorry for that source code!
But yeah, this looks good to me, should work. Many thanks! :)

fabbrimichele:
> I'm so sorry for that source code! But yeah, this looks good to me, should work. Many thanks! :)

Thank you for your great job!

Issue #389, opened 2022-11-04
gbs-control: Will it work with RGBS output from Marconi Specrum Analyser?


brian-mk:
This is a question rather than an issue with GBS-Control...

I have a CRT based Marconi 2380 Spectrum Analyser that provides an RGBS output.
The composite sync output is: horiz: 15.625kHz, vert: 48.2Hz, non-interlaced. i.e. 324 lines.

I have tried a standard GBS-8200 (V4) to convert this to VGA.
Using default GBS-8200 auto settings I get a stable image that is slightly too large in the vertical direction i.e. some lines are missing from the top and bottom of the image.
If I try to reduce the vertical size using the menu buttons, the missing lines become visible but they appear to be out of sync (more so at the bottom).

(See example image below)

Should this 324 line input signal be expected to work with standard GBS-8200 firmware?
If not then if I add an ESP8266 with GBS-Control firmware, could I make this work?
Will it mean making changes to some of the TVIA 5725 registers to suit the 324 line input?

![image](https://user-images.githubusercontent.com/34718037/200066294-09066e6b-961f-477b-a806-725d095555a2.png)

ramapcsx2:
I'd say give it a try. The line count may be odd (forgot the PAL specs.. 325 lines?), but the firmware will try to work with it :)

brian-mk:
I added the ESP8266 module to the GBS-8200 and successfully flashed the GBS-control firmware.

I connected the Spectrum Analyser RGBS output and fed the VGA into a laptop using OBS and a VGA->USB capture device. This is the same setup I used before with the stock GBS-8200. The OBS window remains completely black.

Here is the gbs-control developer mode diagnostic output after powering up the Spectrum Analyser.

====

Activity detected, input: RGB
.*****
no signal
h: 430 v: 647 PLL:0 A:7b7b7b S:07.10.00 H-   I:00 D:05e5 m:0 ht:   0 vt:   0 hpw:   0 u: 96 s: 0 S: 6 W:-71
****************************
no signal
h: 430 v: 647 PLL:0 A:7b7b7b S:07.10.00 H-   I:00 D:05e5 m:0 ht:   0 vt:   0 hpw:   0 u:384 s: 0 S:13 W:-74
************

====

It appears from the diagnostic output that the TVIA 5725 is trying to sync using 647 lines.
Based on the horiz & vert frequencies (15.635kHz & 48.2Hz, non-interlaced) , the analyser output is actually 324 or 325 lines.

[...]

PR #405, opened 2023-01-28
Maintain constant latency with clock generator installed

### LINKED ISSUE #286: Horizontal tear line even with clock generator installed
Especially when playing side scrolling games (Super Mario World in my case) after some play time suddenly a tear line in the lower quarter of the screen appears. This only is visible when actually walking sideways but it is very noticeable and annoying in this case. It disappears once I no longer move.

I have the clock generator installed and the "Sync on Luma" from the SNES is filtered using an LM1881 sync stripper.

If I go to developer options and press "HTotal++" once, then the tear line seems to disappear but ocassionally reappears as a tear line which moves from about the center of the screen to the bottom.

Is this a known problem and is there any reliable fix for this?

nyanpasu64:
This PR continuously adjusts the video output frame rate to keep video output a fixed distance (currently 1/4 frame) behind input, when "FrameTime Lock" is checked and external clock generator is installed. This fixes the bug where when a clock generator is installed, the GBS-Control would initialize input-output latency to a random value, then the latency would slowly drift over time, eventually resulting in a tear line between 0 and 1 frame old images. (Technically the initial latency is still random, but latency is reliably brought to 1/4 frame within seconds.)

The latency control algorithm used is more primitive than I described in https://github.com/ramapcsx2/gbs-control/issues/286#issuecomment-1401230470; we instead use the clock generator to offset the output frequency from the estimated input frequency, by a factor proportional to latency error (clamped to ±0.001 to avoid large frequency offsets). And I'm solely using blocking latency measurements every 100 frames (plus overhead) like your older frame sync code, rather than rewriting the code further to add continuous interrupt-driven non-blocking measurements.

But the results are surprisingly good. If you turn on `FRAMESYNC_DEBUG`, the logs show my code consistently brings latency to a steady-state of within 0.0005 (5 * 10^-4) frames of `FrameSyncAttrs::syncTargetPhase` (90 degrees = 1/4 frame behind input), starting from either console power-on, input resolution changes (240p/480i/480p), or after repeatedly changing output resolution to "480p 576p" to deliberately induce extra latency. I did not test replugging the GBS-C, but I see no reason it wouldn't work well too. I tested primarily at 480p output, but 960p output seems to work fine too.

I also measured video latency (at 480p output) by feeding the input luma signal (through a Y-splitter), and a solar panel at the top of my CRT, into the left and right channels of my my computer's audio line in, then recording at 192 KHz. This showed a latency of just over 4 milliseconds from video input to output, consistent with debug logging.

In "Pass Through" mode, frame sync is unnecessary. My logs show that it is never attempted.

This is a working prototype, but not ready for upstreaming.

## Issues

- [x] The commit history is quite messy, and the diff includes local changes, both refactors I intend to submit a PR for (if deemed acceptable), and machine-specific changes (like `compile_commands.json`, and removing `settingsMenuOLED` to simplify the code because I don't have a screen installed) that should not be merged.
	- I partly added `enum OutputMode`, which is an improvement over the original code, but I'm not confident about the variable names, and is unrelated to this fix. Ideally I'd define a similar enum for `videoStandardInput`, but I don't know the values yet.
		- Interestingly, brian-mk has also mentioned adding enums at https://github.com/ramapcsx2/gbs-control/issues/389#issuecomment-1321164339.
	- I changed the code to echo back every command you send it. I found this useful for figuring out the code that each GUI button triggers, and makes reviewing the logs after the fact more useful, since they tell you which commands you sent in what order, and at what time relative to periodic/delayed events (like sync watcher and frame time lock), input changes, and messages from the code responding to user commands.
	- I moved the existing frame sync interrupt handlers (and associated state) to `namespace MeasurePeriod`, for organization and to prepare for adding new interrupt modes like "measure frame rates and latency in background". I think this change is not harmful, but it's not a clear improvement since I didn't (yet?) implement a new "non-blocking interrupt-driven measurement" mode in a separate namespace like I had originally planned.

- [x] I have not looked into `findBestHTotal` and `applyBestHTotal` yet. I hope this will not produce a vicious cycle of GBS-C continuously increasing htotal to slow down the output frame rate, and my changes continuously increasing the video clock generator frequency to compensate. And after debugging the `FrameSyncManager::reset()` issue, I do not have the energy to learn more code.

- I noticed that the GBS-C's output at 480p is much narrower horizontally than a PC running at 640x480 DMT, or the same GBS-C runing at Pass Through output resolution hooked up to a 480p Wii (almost as wide as PC but offset horizontally). I suspect this is because the GBS-C sets its htotal value much higher relative to the horizontal active time, than the PC standard or the video input. Can this be changed to behave like Pass Through or PC DMT? Increasing the image width through the Picture Control tab/section results in the sides being corrupted.

- When I enter the Developer tab and click "HTotal--" repeatedly (from an initial value of 2345?), the image slowly gets wider. This also calls `FrameSyncManager::reset()`,  which in my initial implementation set `maybeFreqExt_per_videoFps = -1`. This caused frame sync to fail, printing `Error: trying to tune external clock frequency while clock frequency uninitialized!`. (If I press HTotal-- enough times, GBS-C triggers an automatic HTotal resync and prints `HTotal Adjust:  29` in the logs. I'm not sure what codepath triggers this, since I don't have a debugger and didn't go digging.)
	- [x] Do you use a debugger? Should I use the Arduino 1.0/2.0 IDE or VS Code for debugging, or something else?
	- I also got the `maybeFreqExt_per_videoFps < 0` error after manually clicking "Resync HTotal".
	- The error goes away when I set 480p output resolution again.

- The same "clock frequency uninitialized" error happened after power-cycling my Wii. When a video signal reappeared at the same hsync rate as before, this also called `FrameSyncManager::reset()` but not `externalClockGenSyncInOutRate()` to initialize `maybeFreqExt_per_videoFps`.
- I "fixed" this by making `FrameSyncManager::reset()` *not* clear `maybeFreqExt_per_videoFps`. I don't understand the whole state machine, but hopefully this is safe.
	- Now messing with HTotal-- and HTotal++ causes odd things to happen to input latency, where it never converges to near-zero 0.000x even after I press "Resync HTotal" (because we never call `externalClockGenSyncInOutRate` on the new output FPS). Picking an output resolution exits this odd state.
	- I still want to ensure that `FrameSyncManager` would never encounter a mismatched output resolution (I assume switching presets/resolutions always switches `rto->freqExtClockGen`) and `maybeFreqExt_per_videoFps`. To do this, I made `externalClockGenResetClock()` (which sets `rto->freqExtClockGen` when switching presets) call `FrameSync::clearFrequency()` to clear `maybeFreqExt_per_videoFps`, so `FrameSync::runFrequency()` wouldn't operate with the wrong conversion ratio for a different resolution. (`externalClockGenResetClock()` doesn't currently *measure* the output frequency produced by a given clock rate, and I didn't add it since the oscillator and TV5725 may not be stable immediately after programming a frequency.)
		- Now I get a single "clock frequency uninitialized" error when powering my console up, for a brief moment before someone (no clue who, I don't have a debugger set up) calls `externalClockGenSyncInOutRate() -> FrameSync::initFrequency()` which sets `maybeFreqExt_per_videoFps`. I'm fine with this single error, since it goes away (and frame sync begins) when the console remains powered on.
		- Ideally, in external clock generator mode, you wouldn't even call `FrameSync::runFrequency()` before `externalClockGenSyncInOutRate()`. I didn't try changing the code to make this happen.

- **I do not currently have any handling for "freak measurements".** I found that if I use Priiloader to boot straight into the HBC, then exit to the Wii Menu (Health and Safety Screen), there will be a sync phase interruption. Even with "FrameTime Lock" turned off, this usually results in a single `.` (sometimes `2 \n .`) printed to the console. What do numbers, periods, and asterisks mean after the input disappears?
- [x] With my custom "FrameTime Lock" turned on, when exiting the HBC, occasionally (in one attempt out of many) I've seen the *input* FPS be incorrectly estimated as `77.335258`, causing the video clock to be set anomalously high and my VGA CRT monitor to lose sync *after* the Health and Safety Screen appears. I'm guessing my multisync monitor synced to the GBS-C outputting 77 Hz and showed the screen, then when the GBS-C returned to 60 Hz 100 frames later, the monitor went blank before showing the screen again. **Should I fix this in this PR, or make a follow-up one** (since this PR mostly improves behavior with frametime lock turned on, and users can turn it off if they're experiencing problems)?
	- How should I handle this case? Should I save input FPS measurements, and only adjust the value by up to a factor of ±0.01 per sync iteration or 0.05-0.1 total (clamping or ignoring out-of-range values), and let the GBS-C sync watcher(?) handle resetting the output mode when the input switches between 50/60 Hz for a prolonged time?
		- Where should the cached input FPS be stored? Does the source code already save this somewhere? Should we instead derive valid FPS ranges from the input video mode?
		- When should we clear the old FPS measurement or keep it? I'm not interested in repeating the `FrameSyncManager::reset()` debugging adventure without help. Hopefully since we can always rebuild the cached FPS after a clear, it's safe to erase it in both reset() and cleanup(), whenever switching between 15 kHz, 31 kHz, or no signal, even if we don't call `externalClockGenSyncInOutRate` afterwards.
	- Should I take the median of 3 input FPS measurements (with or without rejecting values far from the expected frequency) to adjust output sync? Or if we already have a valid FPS measurement, should we instead take the median of the previous FPS and two measurements?
		- Or should I take two measurements, throw both away and take two more if they disagree by more than 1 FPS, and otherwise average them?
	- Perhaps if sync gets interrupted to the point the GBS-C outputs numbers/dots/asterisks, we should set `delayLock = 0`? I'm not sure how to do that, and it may fail to recognize sync interruptions that *don't* trigger numbers/dots/asterisks but are large enough to tamper with FPS measurements, but I still think it's worthwhile as "defense in depth" (though you have to turn it off when you test FPS measurement rather than playing games).
- If I want to take the median of multiple *sequential* vblank interval measurements, I'd have to rewrite the interrupt code to record the timestamp of multiple sequential interrupts into an array.
	- [x] Is taking multiple *sequential* vblank measurements a good idea since it's faster? Or will that increase the chance of correlated error, since a single erroneous vblank (from noise?) midway through a frame will produce *two* incorrectly low frame time measurements? Should I instead take multiple *independent* vblank interval measurements, like the time from 1-2, then 3-4 (and sleeping from 2-3 without measuring its time), using the current interrupt code?
- Additionally, taking the median of recent frame durations, on top of alternating between input and output frame time measurements, may be tricky to implement in a background interrupt handler. Perhaps you could keep a ring buffer of the last 3 measurements (and possibly the last output), and compute and output the median *within* the interrupt handler (is this too slow?).
	- [x] When writing interrupt-based code, is it safe for loop() to write to a non-volatile global variable before calling `attachInterrupt(fn)`, and `fn` to read from those non-volatile global variables once invoked by an interrupt?

- [x] How likely is it that adjusting the external clock generator within a factor of 0.001 from the frequency derived from the input FPS, will result in wildly incorrect timings (`externalClockGenResetClock()`, see `delay(1)`)? I've never seen it happen, and `externalClockGenSyncInOutRate` makes no attempt to prevent this from happening.

- [x] Will adjusting the external clock generator by a factor of ±0.001 cause monitors to lose sync or change on-screen geometry? My VGA CRT does not lose sync, and I keep thinking I see the on-screen size change, but when I look more closely I can never find it happening at the same time as frequency adjustments.

- If `runFrequency()` repeatedly fails until `syncLockFailIgnore` reaches 0, we call `FrameSync::reset()`. This currently does not reset `maybeFreqExt_per_videoFps`. This *should* be fine? But it won't bring the GBS-C out of a state where `maybeFreqExt_per_videoFps` is persistently unset by mistake.

- I added my own `FRAMESYNC_DEBUG` before noticing `FS_DEBUG`. I don't know the best course of action yet. And Ideally I'd turn off this define before this code is merged (it's useful for debugging, but clutters the logs with constant messages if you're trying to find other messages like resolution changes).

- When I check "Disable External Clock Generator", my code still uses it for latency adjustment until I click "Restart". (I also have to "Restart" after unchecking Disable, to enable the clock generator.) Is this a pre-existing bug?

## Not bugs, but room for improvement

- This does not set latency to a fixed value during start-up. I've seen latencies as high as 0.25 to 0.699 frames above the target, although these are brought down under control in a matter of seconds (and most players are interested in predictable latency during gameplay, not startup menus).

- Should we use a *lower* `FrameSyncAttrs::syncTargetPhase` (video latency)? Should we use a lower video latency when using external clock generator, than with the built-in clock? (I'm not sure if my approach handles 240p/480i transitions better than internal clock and adjusting vblank duration, though it works better in steady-state operation)
	- Ideally users could pick a runtime-configurable target phase (latency in fraction of a frame), and set it lower if they play well-behaved games, and higher when playing games that switch between 240p and 480i (most notably some PS1 games). This would require some refactoring to move phase from a template parameter to a state variable...
	- I've measured up to 0.052 frames of latency offset when switching between 240p and 480i in 240p Test Suite Wii, then waiting for the sync watcher to switch the video output rate and the frame sync to print the resulting change in phase/latency. So setting the latency target to 0.1 frames should be safe even for PS1 games.
	- Oddly sometimes the frame sync notices the FPS change *before* the sync watcher does. The resulting behavior appears harmless?

- Should we frame-sync more often (reduce `FrameSyncAttrs::lockInterval`) (though this blocks the main loop more often)? Should we wait less time after 240p/480i changes before starting frame sync (to stabilize the input latency sooner)? Currently we skip two frames, most likely due to `delayLock`.
	- I do not have the motivation to implement non-blocking frame sync, and want to get this PR (with minimal change from existing code) polished for release first.

- The current code has a separation between `framesync.h` with `FrameSyncManager` and `vsyncPeriodAndPhase` running off periods, and `gbs-control.ino` with `externalClockGenSyncInOutRate` and `getSourceFieldRate/getOutputFrameRate` running off frequencies. My current code straddles the two domains, and I feel the architectural separation makes my changes (and arguably the existing code) harder to understand.

	- For example I reference SerialM, rto, and Si in framesync.h even though they're defined in gbs-control.ino, so clangd treats these variables as undeclared identifiers. And I had to move `#include "framesync.h"` in the main file to below `SerialMirror SerialM;`.
	- Ideally I'd either merge framesync.h into the main file, or move global declarations to a `#pragma once` header and have each file include all headers it uses. And I may or may not try to unify the various methods of getting frame rates.

- I'm scared of how stateful the code is, and how easy it is to perform duplicated work, or (even worse) forget to initialize or update variables. To help find these bugs, I included multiple checks for invalid state, which print an error message. During testing, I've seen multiple cases where `FrameSyncManager::runFrequency` was called after `FrameSyncManager::reset()` cleared `FrameSyncManager::maybeFreqExt_per_videoFps`, before calling `FrameSyncManager::initFrequency()` to initialize it again, and had to think of ways (given my fragmentary understanding of GBS-C's state machine) to preserve or initialize this state when needed, and clear it when invalid.
	- In projects I maintain myself, my current strategy for keeping track is tracking all mutations in an instance of a "StateTransaction" type ([example](https://gitlab.com/exotracker/exotracker-cpp/-/blob/5e32ce834da38f3cc3334163c0ff618fadfec345/src/gui/main_window.h#L238-336)), which sets a bitflag for each piece of modified state, then recomputes all derived state in the non-noexcept destructor (you could also use an explicit commit method to avoid throwing from a destructor).
		- This approach was built primarily around user input, though it also handles periodic events like audio playback ([code](https://gitlab.com/exotracker/exotracker-cpp/-/blob/5e32ce834da38f3cc3334163c0ff618fadfec345/src/gui/main_window.cpp#L1002)). I'm not sure how well this will translate to the GBS-C where you're primarily responding to changes in input video signal, and counting valid/invalid frames before taking action.
	- Another improvement is using `std::optional<T>` and `std::variant` to add "not present" states, and force the interior type `T` to be "always valid" as long as you check that the optional is not nullopt. Then design the program (upfront or through incremental refactoring) around this simplified, more tractable, state management.

(i've spent long enough poring over my code and reading/revising this post over and over...)

Fixes #286.

### DIFF
--- a/framesync.h
+++ b/framesync.h
@@ -614,12 +614,13 @@ public:
         // practice.
         float correction = 0.0038f * latency_err_frames;
 
-        // Most displays can handle the difference between 60 and 59.94 FPS
-        // (ratio of 1.001) with no issue. To avoid compatibility errors, clamp
-        // the maximum deviation from input FPS to this ratio. This is
-        // sufficient as long as fpsInput does not vary drastically from frame
-        // to frame.
-        constexpr float MAX_CORRECTION = 0.001f;
+        // Some LCD displays (eg. Dell U2312HM) lose sync when changing
+        // frequency by 0.1% (switching between 59.94 and 60 FPS).
+        //
+        // To ensure long-term FPS stability, clamp the maximum deviation from
+        // input FPS to 0.06%. This is sufficient as long as fpsInput does not
+        // vary drastically from frame to frame.
+        constexpr float MAX_CORRECTION = 0.0006f;
         if (correction > MAX_CORRECTION) correction = MAX_CORRECTION;
         if (correction < -MAX_CORRECTION) correction = -MAX_CORRECTION;
 
@@ -632,8 +633,9 @@ public:
         // In case fpsInput is measured incorrectly, rawFpsOutput may be
         // drastically different from the previous frame's output FPS. To limit
         // the impact of incorrect input FPS measurements, clamp the maximum FPS
-        // deviation relative to the previous frame's *output* FPS.
-        constexpr float MAX_FPS_CHANGE = 0.001f;
+        // deviation relative to the previous frame's *output* FPS. This
+        // provides short-term FPS stability.
+        constexpr float MAX_FPS_CHANGE = 0.0006f;
         float fpsOutput = rawFpsOutput;
         fpsOutput = std::min(fpsOutput, prevFpsOutput * (1 + MAX_FPS_CHANGE));
         fpsOutput = std::max(fpsOutput, prevFpsOutput * (1 - MAX_FPS_CHANGE));
--- a/framesync.h
+++ b/framesync.h
@@ -21,7 +21,7 @@
 // FS_DEBUG_LED:  just blink LED (off = adjust phase, on = normal phase)
 //#define FS_DEBUG
 //#define FS_DEBUG_LED
-#define FRAMESYNC_DEBUG
+// #define FRAMESYNC_DEBUG
 
 namespace MeasurePeriod {
     volatile uint32_t stopTime, startTime;
--- a/framesync.h
+++ b/framesync.h
@@ -66,6 +66,34 @@ namespace MeasurePeriod {
     }
 }
 
+void setExternalClockGenFrequencySmooth(uint32_t freq) {
+    uint32_t current = rto->freqExtClockGen;
+
+    rto->freqExtClockGen = freq;
+
+    constexpr uint32_t STEP_SIZE_HZ = 1000;
+
+    if (current > rto->freqExtClockGen) {
+        if ((current - rto->freqExtClockGen) < 750000) {
+            while (current > (rto->freqExtClockGen + STEP_SIZE_HZ)) {
+                current -= STEP_SIZE_HZ;
+                Si.setFreq(0, current);
+                handleWiFi(0);
+            }
+        }
+    } else if (current < rto->freqExtClockGen) {
+        if ((rto->freqExtClockGen - current) < 750000) {
+            while ((current + STEP_SIZE_HZ) < rto->freqExtClockGen) {
+                current += STEP_SIZE_HZ;
+                Si.setFreq(0, current);
+                handleWiFi(0);
+            }
+        }
+    }
+
+    Si.setFreq(0, rto->freqExtClockGen);
+}
+
 template <class GBS, class Attrs>
 class FrameSyncManager
 {
@@ -636,8 +664,7 @@ public:
             "Setting clock frequency from %u to %u\n",
             rto->freqExtClockGen, freqExtClockGen);
 
-        rto->freqExtClockGen = freqExtClockGen;
-        Si.setFreq(0, rto->freqExtClockGen);
+        setExternalClockGenFrequencySmooth(freqExtClockGen);
         return true;
     }
 };
--- a/framesync.h
+++ b/framesync.h
@@ -654,9 +654,9 @@ public:
 
         fsDebugPrintf(
             "periodInput=%d, fpsInput=%f, latency_err_frames=%f from %f, "
-            "fpsOutput := %f\n",
+            "fpsOutput=%f := %f\n",
             periodInput, fpsInput, latency_err_frames, (float)syncTargetPhase / 360.f,
-            fpsOutput);
+            prevFpsOutput, fpsOutput);
 
         const auto freqExtClockGen = (uint32_t)(maybeFreqExt_per_videoFps * fpsOutput);
 
--- a/framesync.h
+++ b/framesync.h
@@ -246,10 +246,24 @@ public:
             Serial.println("No");
         }
 #endif
+        #ifdef FRAMESYNC_DEBUG
+        SerialM.printf("FrameSyncManager::reset(%d)\n", frameTimeLockMethod);
+        #endif
+
         syncLockReady = false;
         syncLastCorrection = 0;
         delayLock = 0;
-        maybeFreqExt_per_videoFps = -1;
+        // Don't clear maybeFreqExt_per_videoFps.
+        //
+        // Clearing is unsafe, since many callers call reset(), don't
+        // call externalClockGenSyncInOutRate() -> initFrequency(), then
+        // expect runFrequency() to keep working.
+        //
+        // Not clearing is hopefully safe, since when loading an output
+        // resolution, externalClockGenResetClock() calls
+        // FrameSync::clearFrequency() and clears the variable, and
+        // later someone calls externalClockGenSyncInOutRate() ->
+        // FrameSync::initFrequency().
     }
 
     static void resetWithoutRecalculation()
@@ -297,9 +311,26 @@ public:
 
     static void cleanup()
     {
+        #ifdef FRAMESYNC_DEBUG
+        SerialM.printf("FrameSyncManager::cleanup(), resetting video frequency\n");
+        #endif
         syncLastCorrection = 0; // the important bit
         syncLockReady = 0;
         delayLock = 0;
+
+        // Should we clear maybeFreqExt_per_videoFps?
+        //
+        // Clearing is hopefully safe. cleanup() appears to only be
+        // called when switching between 15 kHz and 31 kHz inputs, or
+        // when no video is present for an extended period of time and
+        // the output shuts off. (cleanup() is not called when switching
+        // between 240p and 480i.) When a new video signal is present,
+        // someone calls externalClockGenSyncInOutRate() ->
+        // FrameSync::initFrequency() to reinitialize the output frame
+        // sync.
+        //
+        // Not clearing is hopefully safe. See reset() for an
+        // explanation.
         maybeFreqExt_per_videoFps = -1;
     }
 
@@ -409,6 +440,10 @@ public:
 #endif
 
         return true;
+    }
+
+    static void clearFrequency() {
+        maybeFreqExt_per_videoFps = -1;
     }
 
     static void initFrequency(float outFramesPerS, uint32_t freqExtClockGen) {
--- a/framesync.h
+++ b/framesync.h
@@ -527,10 +527,14 @@ public:
         bool success = false;
         for (int attempt = 0; attempt < 2; attempt++) {
             // Measure input period and output latency.
-            if (!vsyncPeriodAndPhase(&periodInput, nullptr, &phase)) {
+            bool ret = vsyncPeriodAndPhase(&periodInput, nullptr, &phase);
+            // TODO make vsyncPeriodAndPhase() restore TEST_BUS_SEL, not the caller?
+            GBS::TEST_BUS_SEL::write(0x0);
+            if (!ret) {
                 SerialM.printf("runFrequency(): vsyncPeriodAndPhase failed, retrying...\n");
                 continue;
             }
+
             fpsInput = esp8266_clock_freq / (float)periodInput;
             if (fpsInput < 47.0f || fpsInput > 86.0f) {
                 SerialM.printf(
--- a/framesync.h
+++ b/framesync.h
@@ -527,14 +527,6 @@ public:
             return false;
         }
 
-        if (delayLock < 2) {
-            fsDebugPrintf(
-                "Skipping FrameSyncManager::runFrequency(), delayLock=%d < 2\n",
-                delayLock);
-            delayLock++;
-            return true;
-        }
-
         // ESP32 FPU only accelerates single-precision float add/mul, not divide, not double.
         // https://esp32.com/viewtopic.php?p=82090#p82090
 
--- a/framesync.h
+++ b/framesync.h
@@ -401,6 +401,34 @@ public:
 
         return true;
     }
+
+    // Perform vsync phase locking.  This is accomplished by measuring
+    // the period and phase offset of the input and output vsync
+    // signals, then adjusting the output video clock to bring the phase
+    // offset closer to the desired value.
+    static bool runFrequency()
+    {
+        int32_t period;
+        int32_t phase;
+        int32_t target;
+
+        if (!syncLockReady)
+            return false;
+
+        if (delayLock < 2) {
+            delayLock++;
+            return true;
+        }
+
+        if (!vsyncPeriodAndPhase(&period, NULL, &phase))
+            return false;
+
+        target = (syncTargetPhase * period) / 360;
+
+        // TODO see externalClockGenSyncInOutRate.
+
+        return true;
+    }
 };
 
 template <class GBS, class Attrs>
--- a/framesync.h
+++ b/framesync.h
@@ -504,19 +504,54 @@ public:
         const float esp8266_clock_freq = ESP.getCpuFreqMHz() * 1000000;
 
         // ESP CPU cycles
-        int32_t periodInput;
-        int32_t periodOutput;
+        int32_t periodInput;  // int32_t periodOutput;
         int32_t phase;
-        if (!vsyncPeriodAndPhase(&periodInput, &periodOutput, &phase)) {
-            SerialM.printf("vsyncPeriodAndPhase failed\n");
+
+        // Frame/s
+        float fpsInput;
+
+        // Measure input period until we get two consistent measurements, to
+        // avoid incorrectly guessing FPS when input sync changes.
+        bool success = false;
+        for (int attempt = 0; attempt < 2; attempt++) {
+            // Measure input period and output latency.
+            if (!vsyncPeriodAndPhase(&periodInput, nullptr, &phase)) {
+                SerialM.printf("runFrequency(): vsyncPeriodAndPhase failed, retrying...\n");
+                continue;
+            }
+            fpsInput = esp8266_clock_freq / (float)periodInput;
+
+            // Measure input period again. vsyncPeriodAndPhase()/getPulseTicks()
+            // -> vsyncInputSample() depend on GBS::TEST_BUS_SEL = 0, but
+            // vsyncPeriodAndPhase() sets it to 2.
+            GBS::TEST_BUS_SEL::write(0x0);
+            uint32_t periodInput2 = getPulseTicks();
+            if (periodInput2 == 0) {
+                SerialM.printf("runFrequency(): getPulseTicks failed, retrying...\n");
+                continue;
+            }
+            float fpsInput2 = esp8266_clock_freq / (float)periodInput2;
+
+            // Check that the two FPS measurements are sufficiently close.
+            float diff = fabs(fpsInput2 - fpsInput);
+            if (diff != diff || diff > 0.5) {
+                SerialM.printf(
+                    "FrameSyncManager::runFrequency() measured inconsistent FPS %f and %f, retrying...\n",
+                    fpsInput,
+                    fpsInput2);
+                continue;
+            }
+
+            success = true;
+            break;
+        }
+        if (!success) {
+            SerialM.printf("FrameSyncManager::runFrequency() failed!\n");
             return false;
         }
 
         // ESP CPU cycles
         int32_t target = (syncTargetPhase * periodInput) / 360;
-
-        // Frame/s
-        const float fpsInput = esp8266_clock_freq / (float)periodInput;
 
         // Latency error (distance behind target), in fractional frames.
         // If latency increases, boost frequency, and vice versa.
@@ -541,8 +576,8 @@ public:
 
         #ifdef FRAMESYNC_DEBUG
         SerialM.printf(
-            "periodInput=%d, periodOutput=%d, phase=%d of %d, fpsInput=%f, latency_err_frames=%f, newFpsOutput=%f\n",
-            periodInput, periodOutput, phase, target, fpsInput, latency_err_frames, newFpsOutput);
+            "periodInput=%d, phase=%d of %d, fpsInput=%f, latency_err_frames=%f, newFpsOutput=%f\n",
+            periodInput, phase, target, fpsInput, latency_err_frames, newFpsOutput);
         #endif
 
         // This test should always pass due to MAX_CORRECTION, unless an
--- a/framesync.h
+++ b/framesync.h
@@ -528,8 +528,10 @@ public:
         // Frame/s
         float fpsInput;
 
-        // Measure input period until we get two consistent measurements, to
-        // avoid incorrectly guessing FPS when input sync changes.
+        // Measure input period until we get two consistent measurements. This
+        // substantially reduces the chance of incorrectly guessing FPS when
+        // input sync changes (but does not eliminate it, eg. when resetting a
+        // SNES).
         bool success = false;
         for (int attempt = 0; attempt < 2; attempt++) {
             // Measure input period and output latency.
@@ -552,6 +554,7 @@ public:
 
             // Check that the two FPS measurements are sufficiently close.
             float diff = fabs(fpsInput2 - fpsInput);
+            // TODO switch to relative difference
             if (diff != diff || diff > 0.5) {
                 SerialM.printf(
                     "FrameSyncManager::runFrequency() measured inconsistent FPS %f and %f, retrying...\n",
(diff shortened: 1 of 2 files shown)


ramapcsx2:
commit history: simple: I don't care. At all. The history is a great mess already, how much worse can it get? ;p

findBestHTotal: This comes up in a few variations. Basically, I don't think this is meant to run (or do changes) with the clock gen installed. If it does, then it was for side effects or to keep custom settings in presets.

Do you use a debugger: Nope, feel free to add support, if you want :)

freak measurements: Comes up a in some variations. Basically, if you measure 77Hz on the source, and it's just a 480p Wii, that's wrong :p
It needs to be ignored, another measurement scheduled, maybe the sync situation isn't stable..

when to take VSync period measurements: Experiment and see what you get :)

small clock gen adjustments: The idea is to try and keep the attached display happy, by indeed changing the frequency in very small steps and with some wait time. Whether it works or not depends a lot on the display though. If the display doesn't like it, it will drop sync, go to black screen and resync.

volatile: These are not ideal, but you're guaranteed to read or write atomically on this architecture (no read while it's getting written). Volatiles do not prevent the compiler from reordering or otherwise breaking the intent though. Proper atomic variables would solve it (but are a bit tricky to use).
Basically, I just assumed they work fine, until proven otherwise.

ramapcsx2:
Broadly though, I want to encourage you to try your new code with various sources, and if it seems to work, that's good enough for me. It's going to be an improvement!

nyanpasu64:
https://github.com/ramapcsx2/gbs-control/issues/286#issuecomment-1407641143

> If you can run a few more sources (SNES and PSX for example cover a broad range), the better.

I'm going to test my PR on SNES later today. I don't have a PSX, but running 240p/480i transitions in 240p Test Suite GX showed "up to 0.052 frames of latency offset", which is *well* under the 0.25 frames of target latency. In fact I think even 0.1 frames of target latency would be safe even in games switching between these video modes, but I'm not 100% confident there will *never* be tear lines at that latency.

> If fixes and improvements can be broken up into smaller PRs, then that's good, but sometimes everything is interrelated

Ideally I'd remove the compile_commands.json and restore the OLED menu. At that point, I'll have to decide whether to split out my earlier changes ("Rename serialCommand and userCommand", "Remove commented-out code blocking code folding", `enum OutputMode`, `console.log("GBSControl.ws.onmessage()");` (still not sure what I did in ae28a351f7d9668ba9164b54f824ecd84dc3740d)) into a separate PR.

----

> freak measurements: Comes up a in some variations. Basically, if you measure 77Hz on the source, and it's just a 480p Wii, that's wrong :p

I'm thinking that running the frame lock every 6 frames instead of 100, in a local debug build, will make the incorrect FPS measurements much more reproducible (much higher chance a bad input frame will coincide with measuring input FPS). (Though I should also power off my monitor so I don't strain it with nonstandard refresh rates.)

- [ ] Afterwards I'd test 15/31 kHz transitions and sync phase discontinuities.

Another idea is taking pairs of measurements in a loop until I get two similar vblank duration results (fail after 1-3 non-matching measurements to avoid looping indefinitely), then taking the median of the previous measurement and the two new ones (or if there's no cached measurement, take the first/second/average measurement).

> small clock gen adjustments: The idea is to try and keep the attached display happy, by indeed changing the frequency in very small steps and with some wait time. Whether it works or not depends a lot on the display though. If the display doesn't like it, it will drop sync, go to black screen and resync.

My CRT display doesn't lose sync with 0.1% changes in pixel clock. Perhaps I can release the current code as-is, and if anyone complains of losing sync, I'll consider reducing the maximum frequency deviation then.

nyanpasu64:
Since you squash-merged #406, I rebased and force-pushed on master. There are no code changes.

> maybe I should add a comment `OutputBypass = 10,  // Pass Through` in a subsequent PR? Not sure.

If you're going to squash-merge this PR too, I think it's worth adding this in a separate PR (or you push a commit yourself, to avoid one-liner low-utility PRs) to avoid cluttering the squashed diff. Though if you disagree, I can add the comment here, or drop it entirely (it's not absolutely critical to add).

ramapcsx2:
I've sent you an email :)

nyanpasu64:
Is it possible to eventually implement output-frequency-based latency control, without an external clock generator? I don't know. Even after looking over the register/programming PDFs, I'm not sure what modes the TV5725 has available, to drive the input ADC clock or output DAC clock together or separately, with or without a clock generator, and whether you can fine-tune the output frequency without a clock generator to adjust latency.

I had hopes that `PLL_R/S` and `PLLAD_R/S` would help, but it seems your code already sets those flags, and they control input sampling rate per line rather than output frame rate? Not sure.

ramapcsx2:
The Wii does that sometimes, iirc. It doesn't like the 240p test suite, is all i remember.

nyanpasu64:
Decided to do further testing with a Dell U2312HM LCD monitor, as a stand-in for LCD monitors/televisions.

I connected my GBS-C's inputs to a Wii (via component) and SNES (via RGB), set my GBS-C to output VGA to my monitor at 1080p, then started experimenting with various video modes and signal interruptions. I tested powering on my Wii to HBC in 480p mode (startup random latency/phase), exiting from HBC to Wii Menu (480p randomizing input phase after console power-on), and powering-on and resetting my SNES console (240p startup random phase, and randomizing phase mid-signal).

## Results

It appears that with the GBS-C to LCD signal set to 1080p, the LCD monitor is especially sensitive to changes in input frequency. Anything causing a large increase (haven't seen a decrease work) in GBS-C output frequency will cause the screen to go black for ~1.5 seconds before resyncing. Afterwards, the monitor apparently "re-centers" to the newer frequency, and can freely switch between the old (low) and new (high) frequencies without losing sync. The monitor loses sync again if you increase the frequency *further* past the high frequency without power-cycling, or if you switch to the low frequency and power-cycle the monitor, then raise the frequency again.

With the GBS-C in frequency-based frame-time lock mode (this PR), I've been able to trigger sync loss by sometimes turning on the SNES or Wii, sometimes resetting the SNES or exiting the Wii to System Menu (in frame-time lock), and usually (always?) switching Wii 240p Test Suite between 240p and 480i. (I'm unsure if entering/exiting 240p Test Suite causes a frequency change, as the GBS-C itself loses *input* sync when the Wii switches between 15 and 31 kHz signals.)

Even with frame-time lock disabled, switching the Wii from 240p (263 lines, 59.82 Hz) to 480i (262.5 lines, 59.94 Hz) is always enough to cause my monitor to lose sync. (Oddly switching the other way around, reducing frequency, hasn't yet caused my monitor to lose sync.)

Sidenote: I initially suspected the monitor lost sync because I changed video output frequency too quickly. `externalClockGenSyncInOutRate` included code to gradually ramp the output rate up or down by 1 kHz at a time, and I initially thought 240p-480i transitions did not lose sync. However, removing the frequency ramp did not cause 240p-480i transitions to lose sync, but instead power-cycling the monitor did. Additionally, porting the frequency ramp to `FrameSync::runFrequency` did not fix my monitor losing sync. Even after I switched `FrameSync::runFrequency` to only change output frequency by 0.05% at a time, my monitor still lost sync, but after *two* 0.05% changes rather than one 0.1% change.

I've been able to get my monitor to lose sync *twice in a row* without power-cycling in between, by hard-resetting the GBS-C to create input latency (with External Clock Generator on but FrameTime Lock off), power-cycling my monitor while showing 240p Test Suite (59.82 Hz), switching my Wii to output 480i (59.94 Hz, so `externalClockGenSyncInOutRate` raises the output frequency and breaks sync), then enabling FrameTime Lock (raising the frequency even further to 60 Hz temporarily and breaking sync again). Oddly, after my monitor locks onto sync at 60 Hz, it doesn't lose sync when switching back down to 240p (59.825 Hz input rate, and as low as 59.813 output rate for one iteration to increase latency towards the target).

I decided to check how my monitor behaved at other VGA frequencies. When testing 240p-480i transitions, it seems my monitor can tolerate both input frequency increases *and* decreases when receiving VGA 480p, and can tolerate neither frequency increases *nor* decreases at 720p or 960p. I'm not sure how other LCD monitors behave. (My CRT has no problems with frequency changes, at any resolution I've tested it with.)

(Unrelated observation: 240p Test Suite displaying a wholly scrambled image, that persisted across output resolution changes... perhaps because I had switched resolutions with no input signal, and GBS-C "helpfully" disabled the sync watcher and enabled debug mode, and I hadn't known to turn the sync watcher back on before switching from 480p to 240p. Can I disable this code branch so it doesn't bite me and likely others?)

## Conclusion

Not much we can do about the monitor losing sync. Workarounds:

- [ ] We wouldn't lose sync on console power-on, if we matched the output VGA phase to the input phase when beginning to output a signal. But I don't know how to implement this.
- We would still lose sync on console reset if the input video phase changes, since we have to either adjust the output frequency, or interrupt the output signal just like the input.
- [ ] Should we match input latency sooner (don't wait for `delayLock` in frequency-based sync) after 15/31 kHz transitions, so the output frequency is adjusted immediately (and sync lost immediately) rather than 3+ seconds in? (Displaying video, then losing video multiple seconds later, is quite unexpected behavior.)
	- [ ] Should `externalClockGenSyncInOutRate` call `FrameSync::runFrequency` right away, further reducing latency by a fractional period?
- We cannot avoid changing output frequency upon 15/31 kHz transitions, since the input frequency is fundamentally different. However, we currently lose sync *half a second* after the input mode changes, delaying a stable video half a second further than necessary (and incurring half a second of frequency mismatch and latency drift, which must be corrected for).
	- [ ] To prevent this, we should change output frequency immediately when detecting a line count change. Is this safe to do, and won't trigger false positives? IDK.

I'm going to keep the frequency ramp, because hopefully it prevents the TV5725 from displaying glitched output. In the code without a frequency ramp, one time I saw half a scanline of corrupted output, which coincided with a log message describing changing the frequency by 0.1% instantaneously.

- [ ] Should I revert to 0.1% maximum frequency change at a time? 0.05% is debatably more gradual, but takes around 0.8 seconds longer to lock onto a stable latency, and could theoretically result in latency overshoots if we change frequency too much and too infrequently (but we don't change frequency too far and too infrequently, so we don't overshoot).

ramapcsx2:
Oh, I think you've found a particularly sensitive display then. This is the thing: Nearly every other monitor / TV / collectively "sink" will behave differently. This was annoying to no end when I initially wrote this and dialed in settings..

Priority for where sync drops should be avoided is on 240p/480i switches, as that is a common occurrence when playing games on some systems. The PSX has this bad, so it's a really good test console (has the 240p test suite, too :)).

Next down on the priority list is what happens when the console resets its GPU (PSX again great here), which would happen on a console reset, for example.
I don't think all sync losses can be avoided here, but sometimes, a monitor will endure this fine, just showing a black picture with a few glitches for a bit.

And then furthest down is obvious drastic source timing changes. Sync can be allowed to drop then.

You are asking for what to prefer in some sync loss unavoidable situations:
I think you're on the right track there. If sync will be lost, then it would be best if we get a stable + retimed output back as soon as possible. It will be better to re-time everything at once, provided that the code first determined the new source timings are stable. The PSX is again a good example, where a PAL console will start the GPU in NTSC mode (!) on boot, then switch to PAL a few millis later. A CRT TV will not get too upset by this, but every scaler I know struggles with this a lot, even a RetroTink5X hates this.

Frequency ramping works well on many displays, and I would recommend to stay close to the ramp steps / timings that I had.
They worked across my test gear selection.

nyanpasu64:
> Priority for where sync drops should be avoided is on 240p/480i switches, as that is a common occurrence when playing games on some systems. The PSX has this bad, so it's a really good test console (has the 240p test suite, too :)).

When the input frame rate changes, the output frame rate must change. I think there's flat out *no way* to make my LCD on 720p/960p accept 240p/480i transitions, outside of possibly hacks like transmitting a video signal halfway between the 240p and 480i refresh rates, waiting for the monitor to sync and hoping you don't overflow 0-1 frames of latency, then afterwards transmitting the real refresh rate.

> I think you're on the right track there. If sync will be lost, then it would be best if we get a stable + retimed output back as soon as possible. It will be better to re-time everything at once, provided that the code first determined the new source timings are stable.

The current code is not perfect (probably takes longer than ideal to respond to console resets, due to `lastVsyncLock`), but since I restricted it to a ±0.06% output frequency swing, it does not cause my monitor to lose sync at all. Additionally it takes longer than ideal to respond to 240p/480i switches, but I consider that out of scope for this PR, since I did not write or change that code in this PR (and don't even understand it).

> Frequency ramping works well on many displays, and I would recommend to stay close to the ramp steps / timings that I had.  
> They worked across my test gear selection.

Are you talking about `externalClockGenSyncInOutRate` adjusting the clock generator by 1 kHz at a time? If you think the current formula doesn't need changing, I'll keep it as-is.

## Testing old frame time lock

Are there any TVs or monitors that can't even accept VGA vsync+hsync rate changes as small as 0.06%? (I suspect this is very rare, since you say I have "a particularly sensitive display" and even it can handle 0.075% without problems.)

If so, it may be worthwhile to turn *off* the external clock generator if you want to use frame time lock (bypassing this PR entirely). Afterwards, frame time lock will change vtotal rather than the pixel clock and hsync rate, and some monitors tolerate that better. For example, with internal-only frame time lock, my LCD doesn't shift the image vertically at all in Method 1, and doesn't lose sync when the frame rate changes, even in the ever-picky 960p resolution.

I'd keep external clock generator on for displays like my CRT, which tolerates frequency changes but offsets the image upon line count changes (in *both* FrameTime Lock methods!). Frequency-based frametime lock also has the benefit of converging to a perfectly stable latency, rather than vtotal-based frametime lock causing latency to jitter by significant fractions of a frame in steady-state operation (I think, see below), so I might even use it on my LCD.

- I haven't actually inspected what the old vtotal-based "FrameTime Lock" does using an audio interface (I want an oscilloscope but don't have one). Does it alter the output video signal by 2 scanlines, for a full `lockInterval` at a time?
	- In 480p output mode, in my observation it only enables correction (turns off LED) for 1 iteration at a time, then waits a *long* time before enabling correction again. I suspect slowing down the output rate by 2/525 for a whole 100 frames drastically increases latency (by 200/525 or 0.38 of a frame?!), resulting in a sawtooth latency over time. If accurate, I think this kind of latency variability is quite bad, and bordering on perceptible to twitchy/fighting game players.
		- Ideally I'd decrease `lockInterval` drastically. But that causes the ESP to spend more time blocked and less time ready to respond to user commands or sync changes. And I don't want to further change code that's already outputting a dubiously legal video signal, since I don't have many portable displays I can hook up to my consoles and GBS-C to test for compatibility.
	- On higher output resolutions, adding 2 scanlines per frame is a proportionally lower relative FPS change, and corrections are applied/unapplied more frequently.
- It's useful to uncomment `#define FS_DEBUG_LED`, which causes vtotal-based frame time lock to toggle the LCD when it increases or decreases vtotal. This way, you can correlate vertical image shifts with state changes (if the GBS-C tells you it switched vtotal counts and the image doesn't shift, you know the current "Method: 0/1" works with your monitor).

> Priority for where sync drops should be avoided is on 240p/480i switches

Unfortunately, with "FrameTime Lock" enabled and external clock generator *disabled*, my monitor still loses sync (at 960p) just under a second after the GBS-C input switches between 240p and 480i! This is both the most important case of input frame rate change, and one that "FrameTime Lock" does not fix (and perhaps cannot easily fix).

- This also occurs on master rather than this PR, so it's not caused by my changes, though it's still interesting (and disappointing) to note.
- I found that around 0.8 seconds after a 240p/480i switch, `runSyncWatcher()` hits the `timingAdjustDelay == 0` branch and call `FrameSync::reset`, than 0.1 seconds later the display loses sync. Commenting out `FrameSync::reset()` prevented the screen from going black, but this is not acceptable; after switching from 263 line 240p to 262.5 line 480i, I get a tear line going up the screen (newer image below) every 6 seconds.
	- `reset()` is labeled `sets syncLockReady = false, which in turn starts a new findBestHtotal run in loop()`.
	- Apparently `FrameSync::reset()` sets `FrameSync::syncLockReady = FrameSync::ready()` to false, causing the main loop to call `runAutoBestHTotal() -> FrameSync::init() -> findBestHTotal()`, and change the output frame rate through `applyBestHTotal`.
		- I should look further into `applyBestHTotal`, and see if it provides an easy way to make 480p wider and closer to a PC image or passthrough.
- Could we adapt to input interlacing changes without changing output horizontal frequency? We would have to pick an appropriate horizontal frequency, then change (currently increase only, can we safely decrease too?) output line count by at least a similar amount (1/525) as the input. At 480p output, `FrameSync` already changes output line count by 2/525, but I'm less confident at 1080p output (2/1125?).

nyanpasu64:
## LCD gameplay play-testing

I spent a few days gaming on my Wii outputting 480p component, plugged into a GBS-C upscaling to 960p with clock generator and frame sync enabled, hooked up to my LCD.

- I found 480p/480i transitions (specifically launching GameCube games) to be quite slow, but I'm not sure if this was better on CRT or prior to changes. I'm not too concerned about this, or interested in fixing this.
- Gameplay worked very well, and I have not seen any tearing even at `syncTargetPhase = 36`. I don't know if my LCD adds latency when upscaling 960p to 1080p (this is not a GBS-C problem).
	- Ideally on a 1080p LCD I'd send 1080p from the GBS-C to the LCD. However, in 1080p the GBS-C squashes *all* games into the center 3/4 of each scanline, making it impossible to display widescreen-compatible games in a widescreen LCD.
	- [ ] Is it possible to add a "widescreen 1080p" resolution mode? Or better yet, control whether to pillarbox the output in a separate toggle, independent of the output resolution? (Though I fear that this is difficult to implement due to the TV5725's esoteric video timing register system, which I still don't understand.)

For the later part of my testing, I tried decreasing video latency: I set `syncTargetPhase` to 36 (/360 = 0.1 frame) and reflashed my GBS-C. I am not making this change in this PR, as it requires further testing in 240p/480i transitions, and further changes to be production-ready:

- Reduce `syncTargetPhase` in frequency mode only. (Decreasing target latency in the older vblank mode would change behavior, and possibly even create tear lines.)
- Ideally speed up `lockInterval` in older vblank mode only (to shrink large latency spikes, which I observed even in master), then test the code to ensure it doesn't take up too much time.
- Ideally add a user-selectable target latency, which you can decrease if you don't play PS1 games.

## Low-latency measurements

I then plugged the GBS-C back into my CRT, and checked the resulting latency using a solar panel and audio interface (on Wii Home screen because I'm lazy).

Data at [960p output 0.1 frame target.aup3.zip](https://github.com/ramapcsx2/gbs-control/files/10696602/960p.output.0.1.frame.target.aup3.zip).

- Placing the solar panel on top of the screen, and comparing the beginning of the first scanline, I measured 329 samples of latency, or 1.71 ms. (329 smp / 192000 smp/s * 1000 ms/s)
- Placing the solar panel on the bottom of the screen, and comparing the end of the last scanline, I measured 461 samples of latency, or 2.4 ms.

Results:

- The latency is higher at the bottom of the screen, I think because the output takes longer to scan out each frame (DMT `960p active / 1000 total`) than the input takes to receive it (DMT/CEA-861 `480p active / 525 total`).
- I don't know why even in steady-state, the *frame start* latency is substantially more than 1.666 ms. I didn't check if 480p output would fix this. In any case I don't care enough to investigate, and the latency is good enough for my purposes.

## Conclusion

At this point I'd say this code is reliable and well-tested (though I haven't had other people verify on their devices), and I'm not opposed to merging at this point.

- It works in steady-state operation and console startup on my CRT and LCD (although the 0.06% frequency change limit is a compromise, and slows down startup latency convergence to avoid losing sync on LCD).
- It does not break old vtotal frame sync (method 1 on my LCD still works, does not lose sync, and does not shift the image up and down).
- My LCD still loses sync at 960p (and 720p?) on 240p/480i transitions. It also did on master (before this PR), both with external clock generator enabled, and with clock generator off and frame sync on.
	- I think with a LCD as picky as mine, you cannot avoid losing sync when the input field rate changes, unless you either run a decoupled output frame rate (has tearing), or output a signal which changes vtotal in sync with the input's vtotal (not implemented).

There remain code-level nitpicks; for example I shouldn't have added separate `FS_DEBUG` and `FRAMESYNC_DEBUG` (I think these can be unified, since `FS_DEBUG` mostly affects vtotal sync and `FRAMESYNC_DEBUG` mostly affects frequency sync). But I don't particularly care, since they don't affect users, don't block the critical path of maintenance, and I think they're resolvable easily.

ramapcsx2:
I can say that the 1080p output presets are at the limit of the chip, and indeed compromised. I couldn't get a nice aspect ratio while keeping sharp scaling, so whatever is there now is some form of compromise, not a preferred choice :p 

Latency sounds good. Basically all latencies below 16.6ms / one frame are a great result, and it doesn't matter whether it's closer to 4ms or closer to 2ms. Latency only becomes an issue, once it adds to perceivable lag, which would be in the several frames range.

If you want to preserve more of this work's history, a merge commit is fine.
If it doesn't matter much to you, squash is fine :)

PR #418, opened 2023-02-21
Properly apply Bypass video mode from custom preset

### LINKED ISSUE #416: When saving Pass Through to a preset, no output appears until I select fixed resolution Pass Through
While testing my preset system, I've created a preset which activates Pass Through/Bypass for 480p inputs (with full-sized image), and upscales 15KHz inputs to 480p (unfortunately with horizontally shrunken image) using the full GBS-C processing pipeline. However, after a system reboot, attempting to load this 480p preset produces no output with a 480p input (supposed to activate bypass mode):

```
<reset>
Activity detected, input: Component

2345678
Format change: 3 <stable>
preferencesv2.txt opened
loading from preset slot B: /preset_ntsc_480p.B
ADC offset: R:44 G:43 B:42
clock gen reset: 81000000
ABHT: large diff
retry
Active FrameTime Lock enabled, adjusting external clock gen frequency

preset applied: bypass (custom) for EDTV 60Hz

ABHT: large diff
retry
ABHT: large diff
retry
ABHT: large diff
retry
ABHT: large diff
retry
ABHT: large diff
retry
ABHT: large diff
retry
ABHT: large diff
give up
```

Loading other presets/fixed resolutions, then the Bypass preset again, does not resolve this problem. Only clicking the Pass Through button fixes the preset until the next reboot. Even then, clicking "load preset" still produces odd output:

```
user command 3 at settings source 10, custom slot 66, status 21
preferencesv2.txt opened
loading from preset slot B: /preset_ntsc_480p.B
ADC offset: R:44 G:43 B:42
clock gen reset: 81000000
ABHT: large diff
retry
Active FrameTime Lock enabled, adjusting external clock gen frequency

preset applied: bypass (custom) for EDTV 60Hz

ABHT: large diff
retry
ABHT: large diff
retry
ABHT: large diff
retry
ABHT: large diff
retry
ABHT: large diff
retry
ABHT: large diff
retry
ABHT: large diff
give up
```

Nowhere does it say `pass-through on`, which appears when pressing "Pass Through" normally (printed by `setOutModeHdBypass`). Was saving Bypass into a preset never an intended mode of operation?

I'd say "maybe this bug occurs because loading presets only restores a register dump, and doesn't properly setup Bypass", but I think "restore a register dump" is not strictly true because saving and loading a fixed resolution as a preset shifts 480p sources down (#401).

Reproduced on master e61abf586467d945914e7226cfc101371a87317b.

nyanpasu64:
Fixes:

- Loading and saving custom preset used to produce a preset entry showing the output resolution as Custom, rather than the actual output resolution. This meant that loading and saving a preset was not a no-op.
	- ~~This PR fixes it to save the output resolution *for your current input resolution*. This makes loading and saving a preset a no-op, as long as you saved it in the same *input* resolution as the last time you saved.~~
	- Deferred to a later PR because you can't read hardware (which yields) in a HTTP callback (which must not yield), requiring more invasive code changes.
- Previously loading a custom preset containing Bypass would fail to load properly, if you hadn't already loaded Bypass in the same console power-on.
	- This PR properly reapplies Bypass when reloading custom presets which pass-through sync at the current input resolution.
- Add enum PresetID (incomplete).
- Attempt to better document how userOptions's PresetPreference controls output resolution and runTimeOptions's PresetID.

draft. applies on top of PR #415 (not yet merged), so the github diff view includes irrelevant changes. will be rebased once that's merged.

- [x] if you're comfortable, I could start pushing branches directly to this upstream repo, then I can "stack" PRs on top of each other.

todo:

- [x] merge #415
- [x] combine "Rename OutputMode back to PresetPreference" comment rewrites with commit "Add enum PresetID for derived state"
- [x] add more descriptive multi-line commit messages
- [x] test more

Fixes #416.

### DIFF
--- a/gbs-control.ino
+++ b/gbs-control.ino
@@ -5014,6 +5014,7 @@ void setOutModeHdBypass()
     GBS::PA_ADC_BYPSZ::write(1); // enable phase unit ADC
     GBS::PA_SP_BYPSZ::write(1);  // enable phase unit SP
 
+    GBS::GBS_PRESET_ID::write(PresetHdBypass);
     doPostPresetLoadSteps(); // todo: remove this, code path for hdbypass is hard to follow
     resetDebugPort();
 
@@ -5248,8 +5249,7 @@ void setOutModeHdBypass()
     setAndUpdateSogLevel(rto->currentLevelSOG); // also re-latch everything
 
     rto->outModeHdBypass = 1;
-    rto->presetID = 0x21; // bypass flavor 1, used to signal buttons in web ui
-    GBS::GBS_PRESET_ID::write(0x21);
+    rto->presetID = PresetHdBypass; // bypass flavor 1, used to signal buttons in web ui
 
     unsigned long timeout = millis();
     while ((!getStatus16SpHsStable()) && (millis() - timeout < 2002)) {
--- a/gbs-control.ino
+++ b/gbs-control.ino
@@ -4258,8 +4258,6 @@ void applyPresets(uint8_t result)
                             - we never overwrite rto->presetID = PresetCustomized!
                     - ...
                     - rto->outModeHdBypass = 1; (again?!)
-                    - rto->presetID = PresetHdBypass; // bypass flavor 1, used
-                      to signal buttons in web ui
                 - rto->presetID = PresetCustomized;
         */
 
@@ -5092,6 +5090,12 @@ void setOutModeHdBypass()
 
     GBS::GBS_PRESET_ID::write(PresetHdBypass);
     doPostPresetLoadSteps(); // todo: remove this, code path for hdbypass is hard to follow
+
+    // doPostPresetLoadSteps() sets rto->presetID = GBS_PRESET_ID::read() =
+    // PresetHdBypass. Because we set rto->outModeHdBypass = 1,
+    // doPostPresetLoadSteps() skips assigning rto->presetID = PresetCustomized.
+    // Our caller must assign that manually if needed.
+
     resetDebugPort();
 
     rto->autoBestHtotalEnabled = false; // need to re-set this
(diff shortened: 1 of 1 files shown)


gbs-control.ino
```diff
@@ -7543,8 +7633,8 @@ void updateWebSocketData()
                 case 0x09: // custom
```
guillaumelecerf, line 7633: Maybe:
```suggestion
                case PresetCustomized: // custom
```

nyanpasu64:
gbs-control preset overflow

nyanpasu64:
read corruption after saving a preset:

Screenshot_20230222_030406

refreshing the page fixes the contents read from the slots file. usually the web UI *automatically* reloads the slot list with the correct list a split-second later. this time it didn't, so I was able to take a screenshot.

I was able to reproduce flash corruption on master c4babdd49da0a7d299b3e246d89fb9bb9a5efaaa.

**I now think flash corruption is not caused by the changes made in this PR**, but instead caused by https://github.com/me-no-dev/ESPAsyncWebServer/issues/204:

> the corruption problems seem to have disappeared after I moved the web pages in the firmware image (which is a nice feature of EspAsyncWebServer). In other words, the webserver is not using SPIFFS for static web content anymore. The flash file system is only accessed by the "server.on" functions to read log data and output it in JSON. The same data is also accessed by the logging functions in the main loop, but concurrent access is avoided by means of synchronisation flags. This was not possible when the web pages were loaded from SPIFFS, because the webserver accessed them asynchronously and there was no way to synchronise access with flags.

sounds like it's illegal to even access *different* SPIFFS files concurrently, if you don't manually lock the filesystem. that would explain the separate write and read corruption modes I've observed.

options:

- burn the code down
- "synchronous webserver" (a different library altogether?)
- rewrite the webserver to properly lock the filesystem
- make the esp filesystem multithreaded
- use a different filesystem

nyanpasu64:
marking as draft. you can't yield in a http callback. and reading registers over i2c yields.

fuck AsyncWebServer.

nyanpasu64:
apparently Bypass (both directly applied and from a preset) generates EMI that interferes with WiFi channel 11 communications. This is unrelated to this PR, but makes testing trickier.

PR #440, opened 2023-03-14
Fix warnings in libraries

nyanpasu64:
Fix warning in WebSocketsServer.cpp, and many warnings introduced when updating Si5351mcu in #435.

I cannot reproduce the si5351mcu.cpp bit-shift warnings in Godbolt, but they appear when building gbs-control in Arduino, using both the recommended v2.6.3 esp8266 SDK with old GCC, and the latest 3.1.1 SDK (somehow the folder is named `3.1.0-gcc10.3-e5f9fec` and not 3.1.1).

Godbolt test:

```cpp
#include <stdint.h>

int main() {
    uint8_t x[] = {
        0x0100 >> 8,
    };
    (void)x;
}
```

- [ ] Upstream warning fixes to Si5351mcu?

This also removes a pile of trailing whitespace from the vendored Si5351mcu library, since my editor strips whitespace on saving. I'm not sure whether to upstream this change or not. The main project is free of trailing whitespace, but this library has inconsistent trailing whitespace (but not *consistent* indentation on empty lines).

## Future

There remains one warning from `~/Arduino/libraries/esp8266-oled-ssd1306-master/src/SSD1306Wire.h`, but this one lives in an external library's header included by gbs-control.ino, and cannot be solved without teaching Arduino's build system to expose library header paths using `-isystem` to silence warnings from there.


ramapcsx2:
You could check whether upstream still is active, and if so, ask them.
Otherwise, I'd not hassle them and just strip the whitespaces locally

nyanpasu64:
https://github.com/pavelmc/Si5351mcu was last updated in March 2022 (a year ago), so it's semi active. Should I upstream the warning fixes/silencing only, or the whitespace cleanups as well?

Issue #449, opened 2023-04-01
Random flicker


caiosilva1993:
I'm having small problems with my GBS. Sometimes the image flickers randomly with Nintendo wii emulators in 240p. I did several tests. I disabled and enable the clock generator, disabled and enabled the frame time, checked the cables, enabled the sync button, changed the power source and updated the firmware. But nothing resolved. Any solution?

nyanpasu64:
Did you connect a Wii to the GBS-C using component cables?

In the web UI, on the fourth Settings tab, check Developer Mode. What messages appear in the console to the right when the flickers occur?

caiosilva1993:
I have two nintendo wii, NTSC and PAL, I tested using Ypbpr and RGB, in both cases this anomaly occurs, more specifically when emulating in 240p. Consoles like the SNES this is not observed. The only information that appears in D
![gbs](https://user-images.githubusercontent.com/81496717/229364635-6ab50215-1cbd-4311-bb2c-ca6618f0e5d5.png)
eveloper Mode is a dot. Most of the time, nothing appears.

nyanpasu64:
The image shifts up or down for a frame or so? Maybe you're losing sync. Possibly the issue is related to https://ramapcsx2.github.io/gbs-control/Wiki/Sync-on-Green-Capacitor-Replacements.html. Not sure, this page says you shouldn't replace the capacitors anymore? I don't know if you could have a bad GBS8200 either.

I haven't noticed any sync dropping on my Wii, and I didn't implement sync myself, so I don't know how to debug the problem further.

caiosilva1993:
Here is a example. 
https://user-images.githubusercontent.com/81496717/229400868-72709c6e-a359-47c0-9670-202b37ebb8d1.mp4

ramapcsx2:
In the Web UI, is there a developer option to disable the syncwatcher? If so, please try that, once the console shows a normal image.

caiosilva1993:
I turned off Syncwatcher but without success, the flicker continues.  I believe it is an incompatibility of the Wii in 240p mode with the GBS, because in the ossc this is not observed.  I have another GBS without the genarator clock and curiously I don't observe this happening intensely.  Despite having disabled the clock genarator nothing resolved.  As I said, the SNES works perfectly.

caiosilva1993:
[...]

PR #459, opened 2023-05-03
Fix auto gain on passthrough, properly preserve or reset gain when loading new presets

nyanpasu64:
- [x] Verify correct behavior when loading fixed/custom scaling/passthrough resolutions with auto gain off/on

Tentative test cases:

- With auto gain off, applying fixed scaling should set gain to 0x7B.
- With auto gain off, applying fixed passthrough should set gain to 0x7B.
- With auto gain off, applying custom scaling should apply saved gain.
- With auto gain off, applying custom passthrough should apply saved gain.

And:

- With auto gain on, applying fixed scaling should preserve current gain.
- With auto gain on, applying fixed passthrough should preserve current gain.
- With auto gain on, applying custom scaling should apply saved gain.
- With auto gain on, applying custom passthrough should apply saved gain.

(All verified on Wii 480p input, as of commit fc6acc109fb2bd54941be9127562e8da01d69eb5.)

## Fixes

- Previously enabling auto gain on fixed/custom passthrough would not initialize the gain registers to clipping, and did not enable `DEC_TEST_ENABLE`. This would result in auto gain failing if it was off previously, but working if you switched from scaling to fixed/custom passthrough with auto gain already on.
- Previously, enabling auto gain on a dark scene (so auto gain doesn't reduce gain), then applying a fixed scaling resolution, would set gain to a stale value from the most recent auto gain operation, rather than the current gain. This behavior was unpredictable and confusing.

## Unresolved questions

- [x] An earlier version of this PR would reset auto gain to 0x48 when applying fixed presets. Perhaps it makes more sense to preserve fixed gain across applying fixed resolutions (like current behavior), but reliably and without the chance of loading stale values?
	- Does it make sense for fixed resolutions to reset fixed gain but carry over auto gain? (shrug, preserve existing behavior)
- [ ] Does this increase, decrease, or not affect #449? I haven't gotten any sync drops recently, but this is with quite limited testing, and I haven't swapped capacitors yet.

I'm getting sporadic black screen flashes (so far Super Monkey Ball 2 and Mario Kart Wii, with 480p pixel fix enabled but ineffective) when playing Wii in 480p, with this PR in custom passthrough mode with auto gain enabled. I don't currently know what configurations cause this problem to appear or not.

- [ ] I'll do more testing and see if it still appears with auto gain disabled, then replacing my SoG capacitors once they arrive, and if that doesn't work try 960p upscaling instead, or test *without* this PR (though if that works, then disabling auto gain should've worked already).

### DIFF
--- a/gbs-control.ino
+++ b/gbs-control.ino
@@ -7953,12 +7953,10 @@ void loop()
                 SerialM.print(F("auto gain "));
                 if (uopt->enableAutoGain == 0) {
                     uopt->enableAutoGain = 1;
-                    if (!rto->outModeHdBypass) { // no readout possible
-                        GBS::ADC_RGCTRL::write(0x48);
-                        GBS::ADC_GGCTRL::write(0x48);
-                        GBS::ADC_BGCTRL::write(0x48);
-                        GBS::DEC_TEST_ENABLE::write(1);
-                    }
+                    GBS::ADC_RGCTRL::write(0x48);
+                    GBS::ADC_GGCTRL::write(0x48);
+                    GBS::ADC_BGCTRL::write(0x48);
+                    GBS::DEC_TEST_ENABLE::write(1);
                     SerialM.println("on");
                 } else {
                     uopt->enableAutoGain = 0;
--- a/gbs-control.ino
+++ b/gbs-control.ino
@@ -270,6 +270,13 @@ struct userOptions *uopt = &uopts;
 // remember adc options across presets
 struct adcOptions
 {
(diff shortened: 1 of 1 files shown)


nyanpasu64:
i might've just seen my wii lose sync out the corner of my eye, with auto gain _off_.

i don't know what's wrong. i don't care anymore. maybe my hardware is defective, i'm waiting for my SoG capacitors to arrive. this change is undisputably a bugfix, and will be merged.

PR #465, opened 2023-05-26
Skip overwriting slots.bin if we fail to read the existing contents

nyanpasu64:
Hopefully should guard against slots.bin corruption when the GBS-C crashes.

I haven't actually seen reading slots.bin fail since adding this check. I'll merge it if it happens and prevents corruption.

Do we trust SPIFFS at all?


nyanpasu64:
will just merge it, it can't hurt

PR #472, opened 2023-06-29
Improve 480p passthrough

nyanpasu64:
a subset of `debug-black-screen`, not tested by itself.

- `GBS::HD_HB_ST::write(0x864);` to avoid cutting off 240p test suite full width, not sure if this is already too wide.

not yet:

- [ ] "Switch HD Bypass to always acting like fixed passthrough" (implemented, not included)
- [ ] fix 15khz passthrough with or without double-speed equalization pulses


nyanpasu64:
tested.

Issue #475, opened 2023-07-04
My clock gen doesn't validate according to the doc


brandon-arnold:
Is the clock gen test method up to date in the doc? I've performed a GBS-C AIO build, including the clock mod. The clock mod was done according to both Rama's Github documentation and the GBS-C AIO documentation. But the clock generator test is not passing, for me. 

I should say before anything, I do not trust the AIO PLCC connection with the chip on the GBS-8200. That thing is finicky and I've had to pin it down with zip ties to even get the WiFi module to stay on and the GBS Control settings in the UI to actually change the resolution correctly. But, it is working generally, now, and as I mention below, I've done some fail-safes to ensure the clock mod connections.

The documentation states:

> To test the clock generator is working, open the GBS Control web gui. Go to Preferences and scroll down to Activate FrameTime lock. Press on the FrameTime Lock button and the console will read "Active FrameTime Lock not necessary with external clock gen installed". Video output should be pristine and free of horizontal tearing when high output resolution presets are used.

When I perform this test, the console reads:
```
user command 5 at settings source 1, custom slot 65, status 4
FTL on
Active FrameTime Lock enabled, adjusting external clock gen frequency
```
This indicates the clock mod isn't working, but I've tried so many things to fix it. Has anyone confirmed this is up-to-date and indeed the right way to validate the clock mod?

## Things I've tried

1. Replacing my Si5351 Clock Generator PCB
2. Since the documentation stated 22 uF for the C47/C48 cap replacement, I replacing the 13 uF capacitor that came from GBS-C AIO folks with a 22 uF that measured correctly
3. Manually soldering the SCL and SDA pins directly from the associated pins on the clock board to (a) the MTV230 Myson controller pins 25 and 26 and (b) the SCL and SDA pads on top of the AIO PCB.
4. Verified all said connections with a continuity tester.

YoImLee:
Yes, "Active FrameTime Lock enabled, adjusting external clock gen frequency" means your clock generator is working correctly in more recent versions of the firmware.
 
It's also better to leave it enabled now despite having one installed to take advantage of the new and improved clockgen-based sync lock method.

brandon-arnold:
> Yes, "Active FrameTime Lock enabled, adjusting external clock gen frequency" means your clock generator is working correctly in more recent versions of the firmware.
> 
> It's also better to leave it enabled now despite having one installed to take advantage of the new and improved clockgen-based sync lock method.

[...]

PR #476, opened 2023-07-05
Reference updated clock gen test instructions in Si5351-Clock-Generat…

brandon-arnold:
The instructions on how to test the clock gen are out of date. See #475 where I was confused because of this. According to @YoImLee this should fix the issue.


ramapcsx2:
nice, cheers :)

PR #496, opened 2023-09-03
Added remove presets feature

AlivE-git:
image


nyanpasu64:
I'm happy to see this functionality be added!

Didn't see this before, don't have time to do a full inspection right now, but perhaps extract the string literal used in `slotIndexMap` to a global so it can't go out of sync between the two functions?

(Though it can still go out of sync with `getSlotsHTML` in webui.html, but this isn't really avoidable (though I still kinda want to turn the array in `getSlotsHTML` into a string literal, so you can identify that those character sequences match by grepping, which isn't currently possible).

Did you test that the preset button layout works on desktop and mobile (narrow window width), with and without developer mode enabled?

ramapcsx2:
Oh, same!
Would be great to add, maybe just do your fix and shipit, @nyanpasu64 ? :)

AlivE-git:
Fixed :)
I also checked button layout, there are no problems.

YoImLee:
I'd love to see this being merged. It seems to work well on my personal gbs.

ramapcsx2:
Okay, here goes nothing!

Metaln00b:
@AlivE-git
Can you please also add your changes to the files in the ```gbs-control/public/src/``` (index.ts etc.) folder? Unfortunately, your feature is removed again after the npm install and npm run build. The files you have edited are generated files, if I have understood correctly.

PR #501, opened 2023-09-19
Workaround random hangs when first powering on GBS-C

### LINKED ISSUE #480: Randomly no image when powering on GBS-C
In custom passthrough mode (and IIRC fixed passthrough), connected to a Wii over component, randomly when I power on the GBS and ESP, the GBS-C will fail to initialize and display an image. When I connected to the ESP over serial after the fact, it did not respond to serial commands. I do not know what went wrong during the boot process, as I did not have a USB serial console connected.

I'll have to get a boot log capture before debugging further. I tried plugging in the GBS at various points during Wii startup, but could not replicate the issue.

nyanpasu64:
Fixes #480.

- [ ] Do reliability testing

### DIFF
--- a/gbs-control.ino
+++ b/gbs-control.ino
@@ -7255,6 +7255,10 @@ void ICACHE_RAM_ATTR isrRotaryEncoder()
 }
 void setup()
 {
+    // Wait for power to stabilize, to reduce the risk of the ESP failing to boot until
+    // manually reset (https://github.com/ramapcsx2/gbs-control/issues/480).
+    delay(500);
+
     display.init();                 //inits OLED on I2C bus
     display.flipScreenVertically(); //orientation fix for OLED
 
@@ -9447,7 +9451,7 @@ void handleType2Command(char argument)
             break;
         case 'O':
             // info
-            if (GBS::ADC_INPUT_SEL::read() == 1) 
+            if (GBS::ADC_INPUT_SEL::read() == 1)
             {
                 SerialM.println("RGB reg");
                 SerialM.println(F("------------ "));
@@ -9463,8 +9467,8 @@ void handleType2Command(char argument)
                 SerialM.println(GBS::VDS_USIN_GAIN::read(), DEC);
                 SerialM.print(F("UCOS_GAIN: "));
                 SerialM.println(GBS::VDS_UCOS_GAIN::read(), DEC);
-            } 
-            else 
+            }
+            else
             {
                 SerialM.println("YPbPr reg");
                 SerialM.println(F("------------ "));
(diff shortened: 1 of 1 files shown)


ramapcsx2:
Recommend you do any pin init before your delay, just making sure the IO are in a controlled state. Otherwise, good idea, probably :)

ramapcsx2:
Oh, and what really matters:
This is a WiFi enabled setup. It used to be that the ESP does all the radio setup by itself on boot, which is a huge power hog.
You'll want to look into delaying any WiFi / radio setup, if that's possible. Arduino might make it difficult though.

ramapcsx2:
Pin init: Just the basic IO setup. All the active pads should start in a resting state, where they won't cause trouble or draw current.

But really, if it's for just 500ms, then this PR is fine as it is, those won't matter.

nyanpasu64:
Sorry I haven't looked into this further, I've mostly relied on my power strip to toggle the USB charger's AC input, rather than (flaky) barrel jack plugging. This results in (sometimes) a momentary voltage spike impulse of ±2 volts, then (a second or two later as the MacBook Air USB-C charger starts) a 5-6ms voltage ramp to 5 volts:

DS1Z_QuickPrint31

I haven't gotten any failed startups with this code in place, when toggling the GBS-C charger (and powering on my Wii in sleep mode) through the power strip. Is this good enough to merge?

> Pin init: Just the basic IO setup. All the active pads should start in a resting state, where they won't cause trouble or draw current.
> 
> But really, if it's for just 500ms, then this PR is fine as it is, those won't matter.

I could make the following reorderings (`pin_clk` etc. is for the OLED menu's rotary encoder?), or leave the code as is.

```cpp
void setup()
{
    pinMode(pin_clk, INPUT_PULLUP);
    pinMode(pin_data, INPUT_PULLUP);
    pinMode(pin_switch, INPUT_PULLUP);

    // Wait for power to stabilize, to reduce the risk of the ESP failing to boot until
    // manually reset (https://github.com/ramapcsx2/gbs-control/issues/480).
    delay(500);

    display.init();                 //inits OLED on I2C bus
    display.flipScreenVertically(); //orientation fix for OLED

    //ISR TO PIN
    attachInterrupt(digitalPinToInterrupt(pin_clk), isrRotaryEncoder, FALLING);
```

Also I found out that `void startWire()` calls `Wire.begin()` (which chains to `Twi::init()` with a bunch of side effects including `ets_timer_setfn()/ets_task()` and `pinMode`), then does some pin initialization as well (calling `pinMode` with a *different* argument). I don't think it's safe to move the call to `startWire()` before the delay, because it installs some background tasks that could run. Is it safe to leave this call after `delay()`?

Sidenote: would I have been better off buying a DHO800 now, if I didn't already get a scope a few months ago?

PR #505, opened 2023-10-18
Add platformio support

### LINKED ISSUE #504: Please make the project compatible with Platformio
Unfortunately, it is very difficult to build the project easily.

It not only helps you to develop, but also helps the community to help you to develop.
There are no version numbers of the used libraries mentioned. This always ends in compiler warnings :D

Is there any way I can help you make the project compatible with Platformio?
Is there possibly already a fork for it?

Thanks in advance

Metaln00b:
Currently **not fully tested** platformio version of gbs-control
Build with warnings about SPIFFS (deprecated)

- Add platformio.ini with consideration of CPU frequency 160MHz and flash size 4M1M
    - Add Libraries with version tags (state from ~2019, read from the timestamps of the documentation and its screenshots)
    - In case of upload problems please adjust upload_speed

- Change library ```./src/Websockets*.*``` into ```./lib/WebSockets/*```
- Move library ```si5351mcu``` into ```./src```
- Change includes ```"./src/library.h"``` to ```"library.h"```
- reorder ```static inline void writeBytes(uint8_t slaveRegister, uint8_t *values, uint8_t numValues)```
- predefine ```const uint8_t *loadPresetFromSPIFFS(byte forVideoMode);```

image
v2 Lower Memory (default)

Fixed #504


Metaln00b:
Seems to be working except:

image

I need help :D

ramapcsx2:
So, I'm not sure what to say here..
You know that I don't actively work on gbsc because of other projects, right?
This makes it difficult to judge the impact of big platform affecting PRs like this.
Things need to be tested, there are other coders active that need to verifiy it works for them, etc pp.
Is this worth the effort? I wonder..
Plattformio is an alternative to more vanilla Arduino, but I haven't heard people using it much lately.

Any other's opinion? @nyanpasu64 ? :)

nyanpasu64:
sorry i'm not working actively on (editing) gbs-c right now. I tried porting GBS-C to platformio once in the past, don't remember if I got it working or ran into obstacles. I don't know whether I want to merge a change to the mainline repo, that breaks existing build systems.

If you want, you can fork the project and maintain it yourself, and see if the changes result in a better build/runtime experience.

> Plattformio is an alternative to more vanilla Arduino, but I haven't heard people using it much lately.

In terms of other projects using PlatformIO or not, https://github.com/OpenStickCommunity/GP2040-CE vendors their libraries, https://github.com/JonnyHaystack/HayBox uses platformio I think? I'd say the library management is probably better than Arduino's global libraries or vendoring (at least for supported libraries).

In terms of less positive vibes from the company, https://github.com/platformio/platform-raspberrypi/pull/36#issuecomment-1673289374 platformio seems to be demanding that Raspberry Pi pay them money before they'll merge a Pi Pico PR *not* by the Raspberry Pi company.

ramapcsx2:
Thanks for you thoughts, @nyanpasu64.

PR #506, opened 2023-10-18
Add oled timeout

Metaln00b:
After 10 seconds in main menu, the oled is turning off to prevent burn-in or wear-related darkening


Metaln00b:
Do not accept after merging #505 otherwise cherry-pick

Metaln00b:
Now based on master

Issue #285, opened 2021-10-21
Sync problem with GBS-8200


sv1mne:
Hi, I have a GBS-8200 v4.0 (20091204) which is an excellent video converter!

I have one problem when I use it with a composite video signal with an uncommon sync 19.2 kHz  from an old HP 8924c test set

I can get picture only from YPbPr , but the left part of the picture is cropped, as in the attached image.

Please advise

I thank you in advance

73

Yiannis
SV1MNE
![GBS8200 with HP8924c](https://user-images.githubusercontent.com/34399861/138294249-71355214-4519-4c99-b51d-f5e6d6632b9e.jpg)

GXTX:
I have a similar issue with a arcade console I have (Chihiro). If I directly attach it to my capture card I can shift it manually, but when scaled via GBS the part of the signal is just gone. I've attempted shifting in GBS as well as shrinking and it's still not there.

I don't know the exact sync but it is slightly not standard, older VGA monitors aren't able to sync but most newer ones are.

![image](https://user-images.githubusercontent.com/858612/140593589-0b8d15a0-c60a-4a6d-8c2d-b64d4920ad0e.png)
This isn't so much of a big deal as the games would expect that part of the screen to be overscanned in most cases.

skiphansen:
@ramapcsx2 Hi Robert, my application is the same as the original poster and I had the same results more or less.

Here's a scope capture of the video timing.

![composite_sync](https://github.com/ramapcsx2/gbs-control/assets/4642629/b62195a3-e77c-4543-95c4-39d1b9a48dde)

Just as a back ground the hp8924c was designed primarily for CDMA cell phones which are now long obsolete.  As a result these monsters (59 lbs) have appeared on the surplus market.  Ham radio operators have picked them up over the years for use as general purpose RF test equipment.  They originally sold for > $30k 25 years ago and still command $1K+ used today.

Since they are now over 25 years old they are starting to die and replacement CRT monitors are very expensive if you can find one at all.   

So ... this project might allow us to replace continue to use these excellent pieces of test equipment with "modern" VGA monitors.

Thanks for any help you can offer !

ramapcsx2:
Well, all I can say for pretty sure is that the GBS will be able to handle this, given the right config.
Figuring this out though requires someone to develop it, using exactly this test gear as the source signal.
Some points:
[...]

PR #507, opened 2023-10-22
Add platformio with small impact

Metaln00b:
This is a variant with a smaller impact to support platformio.
This is not fully tested but it runs very well on my odv gbs-c

Fixed #505

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -380,3 +380,9 @@ ASALocalRun/
 
 # MFractors (Xamarin productivity tool) working folder
 .mfractor/
+
+.pio
+.vscode/.browse.c_cpp.db*
+.vscode/c_cpp_properties.json
+.vscode/launch.json
+.vscode/ipch
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,10 @@
+{
+    // See http://go.microsoft.com/fwlink/?LinkId=827846
+    // for the documentation about the extensions.json format
+    "recommendations": [
+        "platformio.platformio-ide"
+    ],
+    "unwantedRecommendations": [
+        "ms-vscode.cpptools-extension-pack"
+    ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "gitdoc.enabled": false
-}
--- a/gbs-control.ino
+++ b/gbs-control.ino
@@ -22,6 +22,10 @@
 #include "SSD1306Wire.h"
 #include "fonts.h"
 #include "images.h"
+
+static inline void writeBytes(uint8_t slaveRegister, uint8_t *values, uint8_t numValues);
+const uint8_t *loadPresetFromSPIFFS(byte forVideoMode);
+
 SSD1306Wire display(0x3c, D2, D1); //inits I2C address & pins for OLED
 const int pin_clk = 14;            //D5 = GPIO14 (input of one direction for encoder)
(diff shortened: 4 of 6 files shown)


kfix:
I've done something similar on my local GBS-C repo a year ago, because I had used VScode+platIO for other esp32 projects and never really used Arduino Studio in the past and didn't want to invest in another toolchain to make some local modifications.

But now I want to try @nyanpasu64's improvements and don't really want to rebase it, so thank you for this PR! I'll give this a try.

skiphansen:
@Metaln00b Thank you!  

I can independently verify that

```
clone https://github.com/ramapcsx2/gbs-control.git
git remote add Metaln00b https://github.com/Metaln00b/gbs-control.git
git merge Metaln00b/add-platformio-small-impact
pio run
pio run -t upload
```
worked beautifully and I'm now playing with the GBS control menu.

I vote for this to be merged, saved me from installing the Ardunio IDE just to build this project.

ramapcsx2:
Hey, the problem is still that other developers need to verify that their workflows aren't broken by this.
A platform change is substantial.

skiphansen:
Absolutely!  I just wanted to add my 2 cents that it worked for me and that I found it to be valuable.

Speaking of value, this is a FANTASTIC project you have here!  Absolutely beautiful !

I'm using it for a different purpose than most same as this guy (https://github.com/ramapcsx2/gbs-control/issues/285)
and I'm having the same problem.  I'll provide more information for that issue later.  

Thanks for the code !!!

nyanpasu64:
If there are multiple people in favor of this change, I'm tentatively in favor of merging it (while tagging the last Arduino-based commit), but we may need to rewrite the docs to avoid confusing new users. However I'm currently dealing with other issues (like getting flagged as under 13 on Discord), so cannot test.

skiphansen:
I didn't look in detail, but there's no reason that adding platformio support should disable / remove Arduino support.  At least not in general.  It should just be an alternate method of building and working on the project.

Metaln00b:
> I didn't look in detail, but there's no reason that adding platformio support should disable / remove Arduino support. At least not in general. It should just be an alternate method of building and working on the project.

I agree with that. I also see that the Web UI in the Docs doesn't seem to be up to date either. We have the ability to maintain our own repos, that's true. I don't want to oppose anyone either. The right to decide is @ramapcsx2 and I respect that!

Friendly opinion:
But I can see that some people tend to be put off. I'm referring to the OLED menu, for example: That was implemented really nicely and confirmed several times that it works, but it hasn't been merged yet.

Since an OLED darkens quite quickly after a while if it keeps displaying the same thing, I have made an adjustment so that the screen turns off after a few seconds in the main menu. Unfortunately, nobody has tested it yet except me. So I can understand if it is not merged.

I also think that if a commit breaks something, it would be possible to jump back one commit, that's what the VCS is there for. But it's also true that if you don't have time to moderate, that's not so great either.

ramapcsx2:
Just keep in mind I can't put any time into this. I'm always busy working on the other projects these days..
So I can not test any of this.
If you guys say it works, then we can merge :)

nyanpasu64:
> I didn't look in detail, but there's no reason that adding platformio support should disable / remove Arduino support. At least not in general. It should just be an alternate method of building and working on the project.

When I tried migrating to platformio (not being familiar with platformio), it would create a new project and copy existing source files to new locations. It seems that this PR adds new build files in parallel with the existing ones, without moving existing source files. If the build works, this is preferable and I'm willing to merge it.

PR #596, opened 2025-02-04
Updated to use ESP Async WebServer v3+ API

dkozel:
The returned parameter values are const and need to be indexed using size_t not the implicit int type.

This might be reverse compatible with the v2 API, but I haven't tested that. I also haven't done testing of every API endpoint, but all of the logical flows look unaffected by making the returned parameter `const`.

I see the large PR https://github.com/ramapcsx2/gbs-control/pull/551 which removes ESP Async WebServer, but hope this small change will be useful to allow users to compile with the up-to-date library. There's also a documentation update for the wiki which should be done as the two ESP Async libaries have moved repos, I'll open a PR and link it to this one.

This was tested with Arduino 2.3.4, ESP8266 board package 3.1.2, ESP Async TCP [0b6de755](https://github.com/ESP32Async/ESPAsyncTCP/commit/243bd5950f1e3b52e83265e6520618ff0b6de755), and ESP Async WebServer [v3.6.2](https://github.com/ESP32Async/ESPAsyncWebServer/releases/tag/v3.6.2).

ESP Async TCP is frozen and will not be further developed according to the maintainer as the ESP8266 is considered mature.

### DIFF
--- a/gbs-control.ino
+++ b/gbs-control.ino
@@ -9542,7 +9542,7 @@ void startWebserver()
             //Serial.print("got serial request params: ");
             //Serial.println(params);
             if (params > 0) {
-                AsyncWebParameter *p = request->getParam(0);
+                const AsyncWebParameter *p = request->getParam((size_t) 0);
                 //Serial.println(p->name());
                 serialCommand = p->name().charAt(0);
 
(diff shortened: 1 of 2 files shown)


dkozel:
Realized that the platformio.ini file needed to be updated with the newer dependencies. I also added the esp12a board as I'm using a NodeMCU v3.

AnvilByte:
I tested this myself and wifi settings isn't being saved, which is a known issue with ESP8266 board package 3.0+
and using 2.6.3 gives me a "error: missing binary operator before token "(" #if __has_include("ArduinoJson.h")" error, so I don't know what's going on

reav89:
I found a fix to the issue with WiFi settings and ESP8266 board package 3.0
I've create a PR #665.
I updated all the libraries to the latest version and was able to compile without any problems.

PR #606, opened 2025-03-04
Turn off OLED instead of screensaver

PBrookfield:
Turns off the OLED instead of displaying the 'Press Any Key' animated message, when the screensaver time period elapses.


PBrookfield:
Sorry, embarrassing bit of ham-fisted github there, just following force-of-habit with pull requests for normal dev work and I'm more of a VSTS rat.  Please ignore!

PR #683, opened 2026-07-18
Add optional compile-time ENABLE_WIFI flag to build without WiFi

settyan:
## What

Adds an optional compile-time `ENABLE_WIFI` flag (default `1`, so existing builds are unchanged) that lets gbs-control be built without WiFi.

Building with `-DENABLE_WIFI=0` — via the new `d1_mini_nowifi` PlatformIO env — leaves out WiFi, the web UI, WebSockets and OTA, drops the ~1 MB embedded web page, and keeps the radio off at boot via `preinit()`. The device is then operated entirely from the OLED menu + rotary encoder.

## Why

The ESP8266 WiFi module is not certified for radio emission in some regions (for example Japan's Giteki / 技適). This gives an opt-in no-WiFi build for those cases. It is not a substitute for radio certification.

## How it is built

- PlatformIO: `pio run -e d1_mini_nowifi` (a bare `pio run` still builds only `d1_mini`).
- Arduino IDE: set `#define ENABLE_WIFI 0` in `gbs-control.ino`, `OLEDMenuImplementation.cpp` and `PersWiFiManager.cpp`.

## Changes

- `ENABLE_WIFI` flag and `#if ENABLE_WIFI` guards around WiFi / web server / WebSocket / OTA in `gbs-control.ino`, the OLED WiFi menu items, and `PersWiFiManager`.
- `SerialM` falls back to plain `Serial` when WiFi is off (reuses the existing `#define SerialM Serial` path).
- `preinit()` keeps the radio off from boot (no-WiFi build only), using the raw SDK so that `ESP8266WiFi.h` stays excluded — any stray `WiFi.*` then fails to compile.
- New `d1_mini_nowifi` PlatformIO env: `-DENABLE_WIFI=0`, excludes `PersWiFiManager.cpp`, prunes the async web libraries.
- `#error` guards for unsupported combinations (no-WiFi + legacy OLED menu, no-WiFi + pinger).

Documentation for the no-WiFi build is proposed separately for the `gh-pages` wiki (Software Setup page), following the project convention of keeping build/usage docs out of the README.

## Limitations (not in scope here)

- Settings that today are reachable only from the web UI (deinterlace mode, FrameTimeLock, line filter, PAL60, external clock, ADC calibration, fine colour/gain), OTA updates, and preset backup/restore are unavailable in a no-WiFi build.
- The ESP8266 SDK radio libraries remain linked; the radio is never started, not removed from the binary. Verify no RF emission on hardware if you rely on it.

## Testing

[...]

PR #684, opened 2026-07-18
Wiki: document building without WiFi (ENABLE_WIFI flag)

settyan:
## What

Adds a "Building without WiFi (optional)" section to the Software Setup page, documenting the new `ENABLE_WIFI` build flag: how to build the no-WiFi variant (`d1_mini_nowifi` PlatformIO env / Arduino IDE), what it leaves out, and its limitations (settings that are web-UI only, OTA, preset backup/restore). It also notes that the ESP8266 SDK radio libraries stay linked, so the radio is never started rather than removed, and that no RF emission should be verified on hardware.

## Ordering

Depends on #683 (the `ENABLE_WIFI` compile-time flag). This page documents a flag that does not exist in `master` yet, so this PR is kept as a **draft** and should be merged only after #683 lands.

### DIFF
--- a/Wiki/Software-Setup.md
+++ b/Wiki/Software-Setup.md
@@ -135,6 +135,23 @@ With the ESP8266 board connected to your computer, and the correct COM port sele
 Do this via __Sketch__ > __Upload__ (hit "Ctrl + U" or by clicking the right arrow in the tool bar).   
 (This could take a while for the first compilation.)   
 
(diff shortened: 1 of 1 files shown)


nyanpasu64:
yeah i'm gonna be unsubscribing from these issues, i haven't used or maintained this project in years and this is ai code, if someone wants to fork go ahead

PR #685, opened 2026-07-22
Add optional compile-time ENABLE_WIFI flag to build without WiFi

settyan:
conversation https://github.com/ramapcsx2/gbs-control/pull/683#issuecomment-5040668697

## Summary

Adds an optional compile-time ENABLE_WIFI flag (default 1, so existing builds are unchanged). Building with -DENABLE_WIFI=0 — via the new d1_mini_nowifi PlatformIO env, or by editing config.h — leaves out WiFi, the web UI, WebSockets and OTA, and drops the ~1 MB embedded web page. The device is then operated from the OLED menu only.

Motivated by regions where the ESP8266 module is not radio-certified (e.g. Japan).

## Validation

Build:
- Vanilla Arduino build (d1_mini) of ENABLE_WIFI=1 is byte-identical between master and this branch (same .bin sha256).
- Arduino and PlatformIO builds both pass, in ENABLE_WIFI 1 and 0.

ENABLE_WIFI=0 (verified on hardware):
- No gbscontrol Wi-Fi network is broadcast (SSID absent from scans).
- gbscontrol.local no longer resolves over AP/STA.
- WiFi entries no longer appear in the OLED menu.

### DIFF
--- a/OLEDMenuImplementation.cpp
+++ b/OLEDMenuImplementation.cpp
@@ -1,9 +1,6 @@
 #define OSD_TIMEOUT 8000
 
-// Per-translation-unit ENABLE_WIFI default (a #define in the .ino does not reach here).
-#ifndef ENABLE_WIFI
-#define ENABLE_WIFI 1
-#endif
+#include "gbs-config.h"
 
 #if ENABLE_WIFI
 #include <ESP8266WiFi.h>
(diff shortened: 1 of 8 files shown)


ramapcsx2:
Cheers @settyan. Wouldn't want to skirt these modern times laws ;)

settyan:
@ramapcsx2 
haha yeah, thanks a lot!!

